PR #6, opened 2020-10-15
Fix docker

SnowNooDLe:
psha-dev is up online


sungeunbae (changes requested):
Currently Travis is configured to run Dockerise.sh this way:
`python scripts/ec2_send_cmd.py "$EC2USER" "$EC2PASSWD" "cd /home/ec2-user/seistech/docker/$BRANCH_NAME;../Dockerise.sh $BRANCH_NAME 5102 5002`

such that it takes branch name and port names. See how it is set up on EC2 and incorporate this design (each branch directory has .env file). This way, you can have a single  Dockerise.sh (without duplicating it for each deployment).

sungeunbae (approved):

PR #8, opened 2020-11-05
GMS Viewer

SnowNooDLe:
# GMS Viewer

1. Implemented GMS_available_IMs so the IM selector is a bit different to IM Selector in Hazard Curve (Seismic Hazard).
2. Database needs to be implemented
Screenshot 2020-11-05 16:58:58

# Instruction

Just like other pages, give a brief guide of how to use this GMS page.

Screenshot from 2020-11-05 16-19-12

## First Plot

The first plot after users compute GMS, default plot is the first index of IM Vector that users chose and using the dropdown, users can see a different plot.

Screenshot from 2020-11-05 16-52-55

## Second Plot

The second plot, ultimately this will go into the first plot as an extra option but not at this point.

Screenshot from 2020-11-05 16-53-03

## Third Plot

Just like the second plot, this plot will go into the fourth plot as an extra option but not now.

The red box is drawn with `DefaultCausalParameters` and can be changed once users update the table. (Pre-GM filtering parameters table)

Screenshot from 2020-11-05 16-53-41

## Fourth Plot

1. The dashed-dotted red lines represent Min/Max (values from `DefaultCausalParams`)
2. The solid red line in VS30 metadata is the default VS30 (From Site Selection tab)

Screenshot from 2020-11-05 16-53-22
Screenshot from 2020-11-05 16-53-31

## Download

Just like any others, it only becomes available once we get the response from Core API. Ideally, the file name will be something more meaningful than just gms.zip but will decide this later on. (Just like the filenames for images - Disaggregation)

## TODO 
1. Some styling may have to be changed.
2. react-plotly.js - based on the chat I had with Brendon last time, plots are more likely there to give a visualisation instead of actually checking every single value or interacting with it, so may not be an issue but some of the plotly.js's functions behaviour is a bit different to each other. (ex: reset axes, double click, drag, zoom). Ultimately, we can disable everything and purely display non-interactive plots which would make life so much better but will see.
3. Check Download data - Something is happening to Download Data, on Postman, we get some responses, on the frontend, we can download the zip but cannot open it.

Screenshot from 2020-11-05 16-45-03


jasonmotha (commented):
Let's update the names from first second and third plot. Otherwise looks good with a quick skim.

claudio525 (approved):
:+1:

SnowNooDLe:
Will open another one when its ready

PR #12, opened 2020-11-17
plotly.js images filename

SnowNooDLe:
# Customization

Update based on Jason's feedback.

Instead of using Station, we use Lat & Lng (with 4 dp)

image

### DIFF
--- a/frontend/src/components/Hazard/SeismicHazard/HazardBranchPlot.js
+++ b/frontend/src/components/Hazard/SeismicHazard/HazardBranchPlot.js
@@ -5,7 +5,7 @@ import { getPlotData } from "utils/Utils";
 import { PLOT_MARGIN } from "constants/Constants";
 import ErrorMessage from "components/common/ErrorMessage";
 
-const HazardBranchPlot = ({ hazardData, im, station }) => {
+const HazardBranchPlot = ({ hazardData, im, lat, lng }) => {
   if (hazardData !== null && !hazardData.hasOwnProperty("error")) {
     const branchHazard = hazardData["branches_hazard"];
     // // Create the scatter objects for the branch totals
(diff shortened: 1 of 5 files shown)


claudio525 (approved):

frontend/src/components/Hazard/SeismicHazard/HazardBranchPlot.js
```diff
@@ -73,6 +73,9 @@ const HazardBranchPlot = ({ hazardData, im }) => {
         useResizeHandler={true}
         config={{
           displayModeBar: true,
+          toImageButtonOptions: {
+            filename: `Branches_Hazard_Plot_${im}_${station}`,
```
sungeunbae, line 77: Just a question. Is there any chance where `im` or `station` contains underscores?

SnowNooDLe: I only played station with CCCC so cannot confirm it but IM, yes, pretty much everything but PGA has underscore.

jasonmotha (approved):
looks good. As claudio mentioned a while back. Station_name is internal to us. So we probably don't need it on the filename but instead maybe lat/lon?

frontend/src/components/Hazard/SeismicHazard/HazardBranchPlot.js
claudio525: I don't think there are any stations that contain underscores, in terms of the IMs with underscores I think just removing the underscore would be fine, i.e. pSA_5.0 becomes pSA5.0
Actually on that, we don't want dots in the name, so for the IMs those should also be changed to p
I.e. this is what I recommend:
`f"Branchs_Hazard_Plot_{im.replace('_', '').replace('.', 'p')}_{station}"`
The above is python code, no idea how to do this in javascript

Also jason's point below is also good, lat  & lon makes more sense

jasonmotha (approved):

PR #14, opened 2020-12-04
NZS 1170

SnowNooDLe:
# NZS1170

NZ Code is now separated, this is no longer part of the Hazard Curve nor UHS. (In terms of API calling).

## NZS1170 Section

We now have a dedicated section for NZS1170 and by setting the location from **Site Selection**, we also get the following default parameters:
1. Distance
2. Soil Class
3. Z Factor
However, only Soil Class and Z Factor are visible on the frontend. **Z Factor** is the user input with float, **Soil Class** is the dropdown with five options.

image

## NZS1170 API Call

There are three different ways of calling NZCode API.

1. Compute Hazard Curve
2. Compute UHS
3. Compute NZS1170

### 1. Compute Hazard Curve

Just like the image below, when **Hazard Curve**'s compute button gets clicked, it sends the first request to get Hazard data then send another request to get the NZ Code data.
image

### 2. Compute UHS

Just like Hazard Curve above, after the app sends a request to get UHS data, it sends another request to get NZ Code data.

### 3. Compute NZS1170

Compute button is disabled until one of **Hazard Curve** or **UHS** gets computed. When one of those get computed, users now can update **Soil Class** and/or **Z Factor** then click Set button to set it then the compute button. It will then update **Hazard Curve** and/or **UHS** NZ Code data without them to be recomputed.

The logic on this compute button is basically:

1. We need to have at least one plot, either **Hazard Curve** or **UHS**
2. **Soil Class** and/or **Z Factor** needs to be set by the Set button (Without clicking set, will not update value)
3. New **Soil Class** and/or **Z Factor** should not be equal to the current NZ Code's parameters.
(For instance, we compute the Hazard Curve with default parameters which are Soil Class - B, Z Factor - 0.3. then until we change either Soil Class and/or Z Factor to something else, Compute will stay disabled as we already used B and 0.3 to get NZ Code data.)

##### Hazard Curve Compute with D - soft or deep soil & 0.3

image

##### Update Hazard Curve's NZ Code via NZ Code section with A - rock & 0.9

image

##### UHS Compute with D - soft or deep soil & 0.3

image

##### UHS's NZ Code via NZ Code section with A - rock & 0.9

image

## NZ Code visibility

With the checkbox, we can toggle on/off the NZ Code plot's visibility.

#### With NZ Code
image

#### Without NZ Code
image

#### With NZ Code
image

#### Without NZ Code
image


jasonmotha (approved):

claudio525 (changes requested):
See https://github.com/ucgmsim/seistech/pull/335 before this can be merged

sungeunbae:
Excellent documentation

claudio525 (approved):

PR #29, opened 2021-01-13
MariaDB

SnowNooDLe:
# MariaDB with Docker

We now have a MariaDB implemented. It is there for the following reasons,

- Keep track users activity (What they have done, e.g., Hazard Compute, UHS Compute, etc...)
- Allocate project(s) for each user - will be implemented properly in the future.
  - Reason behind this is we have a list of projects but they aren't for everyone, some users will have more/less than the other and we will use this DB to pull whats available projects for each user. 
  - For now, we only have two IDs with one potential alpha tester so keep what we currently have on our Project API that returns every possible Project IDs (gnzl and nzgs_pga)

## User table

I wrote a python script that we can add our current exist Auth0 users to our DB.

image

## History table

History table will record the following things:

- History ID - Primary key and we can use this key to track the certain event's/activity's attributes from **History_Request** table.
- User ID - To identify who did what.
- Endpoint - The action they did.
- Date - NZ timestamp as our clients will be NZ based.

image

## History_Request table

image

Based on the History ID, we know what users sent with. So in the future, if something goes wrong from the users' side, we can also track the values they put and we can see what went wrong.

image

image

# Minor change

As requested, we now have V(subscript)S30

image

image

However, not in the following areas,

image

image

As they are not an HTML element, a string. I could possibly look into this at some stages but assumed that they are not in high priority.


sungeunbae:
love the documentation!

middleware/middleware/add_exist_users.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    # List of dictionaries
+    user_list = resp.json()
+
+    user_id_list = []
+
+    # We want to store an actual id that comes after auth| or google| so split string by |
+    for dic in user_list:
+        for key in dic:
```
claudio525, line 55: I would rewrite this slightly, something like
```
for user_dict in user_list:
    if "user_id" in user_dict.keys():
         user_id_list.append(user_dict["user_id"].split("|")[1])
    else:
        print(f"WARNING: No user_id found for user_dict {user_dict}")
```
Or
```
user_id_list = [user_dict["user_id"] for user_dict in user_list if "user_id" in user_dict.keys()]
assert len(user_id_list) == len(user_list)
```

middleware/middleware/add_exist_users.py
```diff
@@ -0,0 +1,79 @@
...
+    res = conn.getresponse()
+    # Convert the string dictionary to dictionray
+    data = json.loads(res.read().decode("utf-8"))
+
+    return data["access_token"]
+
+
+def get_users():
+    """Get all users"""
+
```
claudio525, line 42: Unnecessary empty line (IMO), run black over this file?

middleware/middleware/models.py
```diff
@@ -0,0 +1,58 @@
...
+    project_id = db.Column(db.Integer, primary_key=True)
+    project_name = db.Column(db.String(100))
+
+    def __init__(self, name):
+        self.project_name = name
+
+
+class User(db.Model):
+    user_id = db.Column(db.String(100), primary_key=True)
+    histories = db.relationship("History", backref="owner")
```
claudio525, line 24: Just `history`? Probably shouldn't take my word on this though

middleware/middleware/models.py
```diff
@@ -0,0 +1,58 @@
...
+    # To create a connection between Project and available_projects table
+    # By using lazy attribute with dynamic, instead of getting all the data to view, get a query to filter out
+    projects = db.relationship(
```
claudio525, line 27: Also is there any reason for defining/loading these relationships (i.e. histories and projects) here, maybe if we used these anywhere, but I don't think we are/will?

middleware/middleware/server.py
```diff
@@ -47,10 +71,91 @@ def handle_auth_error(ex):
...
+    """We store Auth0 id to DB so no need extra step, just pull sub's value which is the unique user_id"""
+    token = get_token_auth_header()
+    unverified_claims = jwt.get_unverified_claims(token)
+
+    user_id = unverified_claims["sub"].split("|")[1]
+
+    return user_id
+
+
+def recored_history(endpoint, query_dict):
```
claudio525, line 84: rename to `write_request_details`?

claudio525 (commented):
Looks good, just a couple of minor things. 
Also thoughts on adding email address to user table to make it possible to determine the person from the user table (currently one has to look at Auth0)? This would be a later change though I think (if we decide that we want to add that). @jasonmotha

jasonmotha:
I think leave email address out for now. We probably will end up adding some more user details to the table. But might as well only add them as need them. 

Ideally, we won't be looking at the db directly but have queries/lookups/reports so we don't have to worry about complete DB tables.

middleware/middleware/add_exist_users.py
SnowNooDLe: I will use the first one as that makes more sense to me for now with my level of Python.

middleware/middleware/models.py
jasonmotha: I think that defining them in the class means that sqlalchemy can do some optimisations. But I'm guessing.

You can probably define similar relations in SQL itself.

claudio525: Not sure I expressed what I mean very well above haha. Lets discuss Monday, and I'll explain what I mean.

claudio525 (approved):
👍

jasonmotha (approved):

PR #34, opened 2021-01-19
Fix the Project's UHS plots and update Project/Hazard's UHS plotly.js label

SnowNooDLe:
# Fix the Project's UHS plots and update Project/Hazard's UHS plotly.js label

UHS Plotly.js plots now include both return period and exceedance rate information in a label.
Return Period is displayed in 4 SF.
Exceedance Rate is displayed in 3 SF.

## Hazard Analysis - UHS Plot

For Hazard Analysis, users put the annual exceedance rate(For now but will have two different inputs, 1. Annual exceedance rate, 2. Return Period) and it will contain `.` in input as we only accept the rate between 0 < x < 1. This case, plotly.js will use either 3SF or 4SF to their label(depends on whether its Rate or RP) to display as they are float/double.

image

## Project - UHS Plot

Whereas for Project, users choose from the given Return Periods and they are integer(for now).

In this case, we do not have to apply any significant figures so display an actual Return Period. But we are adding extra information in a label, Rate, so applying 3SF to the rate to follow the conventions.

image

## Fixed to display selected RPs

Also, it is now fixed to display only the selected RPs plots as we used to plot every possible rate regardless of what users chose.

image
image

### What it used to look like

Screenshot 2021-01-20 11:26:23

### What it looks like now

Screenshot 2021-01-20 10:52:42

### DIFF
--- a/frontend/src/components/common/UHS/UHSPlot.js
+++ b/frontend/src/components/common/UHS/UHSPlot.js
@@ -1,7 +1,12 @@
 import React from "react";
 import Plot from "react-plotly.js";
-import { getPlotData } from "utils/Utils.js";
-import { PLOT_MARGIN, PLOT_CONFIG } from "constants/Constants";
+import { getPlotData, renderSigfigs } from "utils/Utils.js";
+import {
+  PLOT_MARGIN,
+  PLOT_CONFIG,
+  APP_UI_SIGFIGS,
+  APP_UI_UHS_RATETABLE_RATE_SIGFIGS,
+} from "constants/Constants";
 import ErrorMessage from "components/common/ErrorMessage";
 
 import "assets/style/UHSPlot.css";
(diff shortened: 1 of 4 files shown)


frontend/src/components/common/UHS/UHSPlot.js
```diff
@@ -17,28 +22,58 @@ const UHSPlot = ({ uhsData, nzCodeData, showNZCode = true, extra }) => {
...
+      let displayRP = (1 / Number(curExcd)).toString();
       scatterObjs.push({
         x: curPlotData.index,
         y: curPlotData.values,
         type: "scatter",
         mode: "lines",
         line: { color: "blue" },
-        name: `${curExcd}`,
+        name:
+          displayRP.indexOf(".") == -1
```
jasonmotha, line 65: Maybe break this down into a function as it's called twice? Good explanation comment.

I presume that the constants is recycled from the other places you are rounding?

SnowNooDLe: Yeah, the function should work and yes, those constants are used in Hazard Analysis UHS's table as well.

claudio525 (approved):

jasonmotha (approved):

claudio525 (approved):

PR #36, opened 2021-01-21
Cross-check DB and Project API

SnowNooDLe:
# Cross-check DB and Project API

Instead of calling a Project API straightaway, we now do cross-check.
By reaching the following endpoint, it does some job behind.

```python
# Site Selection
@app.route("/projectAPI/ids/get", methods=["GET"])
@requires_auth
def get_available_project_ids():
    return get_available_projects()
```

`get_available_proejcts` is basically doing the following jobs behind:

1. Get the Auth0's unique id from the logged-in user's access token. (As our DB stores this ID)
2. With the unique id, get available projects in a list. (This will be a list of **Project ID** as our Projet table stores Project ID, not a name.
3. Then it hits the Project API's endpoint to get available Project IDs. In a form of a dictionary/object with the following shape: `{project_id: {name: project_name}}
4. Compare those two items (one list and one dictionary) to compare the `Project ID` then we create an object in the form of `{project_id: project_name}` -> It will crate a list based on the UserDB as users can only access to the **Project IDs** that they have permission to access.
5. Send a dictionary in JSON format to the Frontend to deal with.

```python
def get_available_projects():
    """Do cross-check for the projects.

    It finds available projects from the DB (Available_Project that contains user_id and project_name).
    After we get all the existing projects from the Project API.
    Then we compare [Available Projects] and [All the Existing Projects] to find the matching one.
    """
    # Finding an user_id from the token
    user_id = get_user_id()

    # Get all available projects that are allocated to this user.
    available_project_objs = (
        Project.query.join(available_projects_table)
        .filter((available_projects_table.c.user_id == user_id))
        .all()
    )

    # Create a list that contains Project IDs from DB (Allowed Projects)
    available_projects = [project.project_name for project in available_project_objs]

    # Get a list of Project IDs & Project Names from Project API (Available Projects)
    # Form of {project_id: {name : project_name}}
    all_projects_dicts = proxy_to_api(request, "api/project/ids/get", "GET").get_json()

    # Create an dictionary in a form of if users have a permission for a certain project
    # {project_id: project_name}
    all_projects = {
        api_project_id: api_project_name["name"]
        for api_project_id, api_project_name in all_projects_dicts.items()
        if api_project_id in available_projects
    }

    return jsonify(all_projects)
```
## What it looks like now

Instead of displaying gnzl, it now says "Generic New Zealand Locations" which is a more readable label than gnzl. Also, from the photos below, users with different permission will see different options.

Screenshot 2021-01-21 14:47:51

Screenshot 2021-01-21 14:47:22

## Frontend Updated

To make this happen, I had to make some changes. For instance, the following function:
React-select (Library we use for a dropdown) only accepts the Object form for its options. We used to create options like `{value: key, label: key}` as most of them are already read friendly. However, Project ID is a bit different, value is a lot shorter which is not user friendly but we use that unfriendly value as parameters to send to either Core API or Project API. So we need to keep this value but also a friendly label to display in a dropdown, hence, created this function just for Project IDs which displays more user-friendly label and send a right value that we used to send to Project API.

```javascript
export const createProjectIDArray = (options) => {
  let selectOptions = [];

  for (const [key, value] of Object.entries(options)) {
    selectOptions.push({ value: key, label: value });
  }

  return selectOptions;
};
```


middleware/middleware/server.py
```diff
@@ -126,29 +129,43 @@ def write_request_details(endpoint, query_dict):
...
-        )
+        .filter((available_projects_table.c.user_id == user_id))
         .all()
     )
 
+    # Create a list that contains Project IDs from DB (Allowed Projects)
     available_projects = []
 
-    for project in available_project_ids:
+    for project in available_project_objs:
```
claudio525, line 152: You should use a list comprehension for this (see https://www.w3schools.com/python/python_lists_comprehension.asp), they make simple for loops like this easy one-liners.
`available_projects = [project.name for project in available_project_objs]`

middleware/middleware/server.py
```diff
@@ -126,29 +129,43 @@ def write_request_details(endpoint, query_dict):
...
-    return jsonify({"project_ids": available_projects})
+    # Get a list of Project IDs & Project Names from Project API (Available Projects)
+    # Form of {project_id: {name : project_name}}
+    all_projects_dicts = proxy_to_api(request, "api/project/ids/get", "GET").get_json()
+
+    # Create an dictionary in a form of if users have a permission for a certain project
+    # {project_id: project_name}
+    all_projects = {}
+
+    for db_project_id in available_projects:
```
claudio525, line 163: There is also an equivalent one for dictionary, called dictionary comprehension (see https://www.geeksforgeeks.org/python-dictionary-comprehension/), which again can be used to make this whole thing a tidy one-liner.
```python
all_projects = {cur_id:cur_name for cur_id, cur_name in all_project_dicts.items() if cur_id in available_projects}
```

claudio525 (approved):

PR #39, opened 2021-01-27
Fix EA site selection 0.000 issue

SnowNooDLe:
# Fix EA's Site Selection 0.000 issue

The code we used to set the default lat&lng, either the value from `.env` or an empty string when it's not available.

```javascript
export const DEFAULT_LAT = process.env.REACT_APP_DEFAULT_LAT || "";
export const DEFAULT_LNG = process.env.REACT_APP_DEFAULT_LNG || "";
```

However, we kept getting 0.000 instead of an empty input field with a hint:

image

The solution was a bit simple, (even though it took me a while to figure out)

```javascript
// For MapBox as using above two will send request everytime users change its value
  const [mapBoxCoordinate, setMapBoxCoordinate] = useState({
    lat: CONSTANTS.DEFAULT_MAPBOX_LAT,
    lng: CONSTANTS.DEFAULT_MAPBOX_LNG,
    input: "input",
  });
```

We used to set the default MapBox coordinate with `input:"mapbox"` and the issue was from here.

```javascript
value={
                  inputSource.lng === "input"
                    ? localLng
                    : Number(localLng).toFixed(4)
                }
```

Because the way we set the `value` for a TextField is based on checking the input source and if it's from "input", we put whatever the user puts (e.g., a user types -43.53123123124123124234, we display them all) but if it's from "MapBox", we only put up to 4DP in input fields (e.g., a user clicks on the map and the coordinate is -43.34123123124124123, then we only display -43.3412).

Hence by setting this MapBox with default coordinates, telling the app that input is from MapBox,

image

and above code is what they've been doing and the reason it was displaying 0.000 instead of nothing.

It now displays nothing but a hint for Site Selection inputs.

image

## Quick bug fix - Changing an uncontrolled input to controlled.

So this issue occurred while I was trying to fix the 0.000 issues and it happened because some approaches I did. They worked but it set those values to `undefined` which is not an ideal approach. However, if it was an empty string like the photo below, then it's an **empty string** which is valid but `undefined` is not. Hence. the `undefined` is an uncontrolled input and when users either type or click the MapBox, now it has a value in `String` which is a controlled value.

So this error is gone now as we properly managed to set the default values to an empty string, not `undefined` and we are changing from `String` to `String`.

By fixing this issue, learnt what actual uncontrolled & controlled values are.
 
image

## Minor changes - Naming

### DIFF
--- a/frontend/src/context/GlobalContext.js
+++ b/frontend/src/context/GlobalContext.js
@@ -29,7 +29,7 @@ export const Provider = (props) => {
   const [mapBoxCoordinate, setMapBoxCoordinate] = useState({
     lat: CONSTANTS.DEFAULT_MAPBOX_LAT,
     lng: CONSTANTS.DEFAULT_MAPBOX_LNG,
-    input: "MapBox",
+    input: "input",
   });
 
   // For IMs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - If IM does not have one starts with `pSA`, the UHS section will be disabled.
 
+### Fix EA Site Selection 0.000 issue - ([PR #39](https://github.com/ucgmsim/seistech_psha_frontend/pull/39))
+
+- Fixed an input fields issue, displaying 0.000 instead of a hint even with no default values.
+
 ## January 22, 2021
 
 ### HOTFIX - Flask-SQLAlchemy isolation level config - ([PR #37](https://github.com/ucgmsim/seistech_psha_frontend/pull/37))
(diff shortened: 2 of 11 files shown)


frontend/src/context/GlobalContext.js
```diff
@@ -29,7 +29,7 @@ export const Provider = (props) => {
   const [mapBoxCoordinate, setMapBoxCoordinate] = useState({
     lat: CONSTANTS.DEFAULT_MAPBOX_LAT,
     lng: CONSTANTS.DEFAULT_MAPBOX_LNG,
-    input: "MapBox",
+    input: "input",
```
claudio525, line 32: Is there something missing in this PR? This is the only actual change I see, everything else is just renaming?

claudio525 (approved):

jasonmotha (approved):

PR #42, opened 2021-02-09
GMS Viewer

SnowNooDLe:
# GMS Viewer

image

After setting the location, GMS tab will be enabled.

## GMS Inputs

There are four input fields that users can use.

1. IM Type
2. IM Level/Exceedance Rate
3. IM Vector - Multiple options possible
   - It shares the same list of IMs to IM Type except, it will filter out the chosen IM Type. 
4. Number of Ground Motions
 
image

## Example inputs

image

## Advanced tab

With users input(the image above), it will call two Core APIs to get the following details:

1. Pre-GM Filtering Parameters - takes about 1.2 minutes
2. Weights - Instant

image

image

Leave **Database** empty and use 1 as a default value for **Replicates**
image

## GMS Compute

It takes about 40secs to get computed data.

image

Default plot will be always the first chosen IM vector.

image

image

image

image

image

## TODO

1. Polish the code.
2. Add spinner to notify users while fetching `Pre-GM Filtering Parameters` in the advanced tab.
3. Better plotly.js's label, like one in feedback_update branch.


frontend/README.md
```diff
@@ -594,6 +594,120 @@ After users have done the following steps:
...
+## Application
+
+### Hazard Analysis
+
+#### Site Selection
+
+##### Ensemble
+
+If it is in DEV version, you see two options, **v20p5emp** and **v20p45sim**.
+But in EA and PROD.
```
jasonmotha, line 606: unfinished sentence?

jasonmotha (dismissed):
Looks good. Most of it goes over my head though.

jasonmotha (approved):

claudio525 (approved):

PR #44, opened 2021-02-10
Edit User UI

SnowNooDLe:
# Edit User UI

Simple UI for SeisTech team to make life better.

image

## User dropdown

The list is done by using Auth0's endpoint to grab all the existing users.

image

## Project dropdown

Project dropdown will show the list of projects that can be added to the chosen user.

For instance, Brendon will have all five projects but Tom has nothing that can be added.

This is because we compare the list of projects in our DB, `Available_Projects` table to the projects that are returned by the ProjectAPI.

As you can from the image right below, in my localdev DB, it only has five rows for Tom. 

image

From the image above, localdev DB doesn't have Brendon's available projects, hence we can allocate all five projects to Brendon if we want.

image

Whereas for Tom, I already allocated all five projects to myself, hence no options available.

image

## Workflow

1. Select a user from the User dropdown.
2. Select project(s) you want to allocate to the chosen user.
3. Click the Allocate Project button to allocate.

### 1. Select a user from the User dropdown

Remember, the list is coming from the Auth0, not from our DB, hence there may be a case where the DB does not know the chosen users information yet.

For instance, we create an account for Robin & Sarah, then Auth0 knows but our DB until we update. This case, you will still see Robin & Sarah's account under User dropdown.

### 2. Select project(s) you want to allocate to the chosen user.

Continue from 1, if the chosen user is not in the DB, then you will see every project that is currently in the ProjectAPI as the DB does not have any information of the chosen user due to new creation.

### 3. Click the Allocate Project button to allocate

Finally, trying to allocate projects to a new user will do the following steps.

1. Add a chosen user to the DB first as the user is not existing in the DB.
2. Then allocate chosen projects to the chosen user to the DB, available_projects table.

This happens the same to the project if it is not in the DB but exists in the Project API. This process will update the DB then allocate to the chosen. user.

## Examples 1 - User does not exist in the DB yet

User is not there
Screenshot from 2021-02-11 10-24-18

But the user is in the list as it is communicating to the Auth0, not the DB
Screenshot from 2021-02-11 10-24-38

Successfully added the chosen projects to the chosen user.
Screenshot from 2021-02-11 10-24-48

Now the DB also has information about the chosen user.
Screenshot from 2021-02-11 10-25-03

Also, the `available_projects` table is updated.
Screenshot from 2021-02-11 10-25-07

Hence, we only have one project to show as this is the only project that can be added to Tom.
Screenshot from 2021-02-11 10-24-56

## Examples 2 - Project does not exist in the DB yet

Imagine the DB only has four projects.
Screenshot from 2021-02-11 11-12-43

Also in the `available_projects` table, Tom only has permission to four projects.
Screenshot from 2021-02-11 11-12-48

However, ProjectAPI says there are five projects, not four, hence the dropdown will show the fifth option which can be added to Tom.
Screenshot from 2021-02-11 11-14-05

Screenshot from 2021-02-11 11-14-10

The chosen project did not exist in the DB, updating the DB.
Screenshot from 2021-02-11 11-14-13

The DB is updated for both `Project` and `Available_Projects` table.
Screenshot from 2021-02-11 11-14-23
Screenshot from 2021-02-11 11-14-28


middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+    requested_user_id = data["user_info"]["value"]
+    requested_project_list = data["project_info"]
+
+    for project in requested_project_list:
+        add_available_project_to_db(requested_user_id, project["value"])
+
+    return "DONE"
+
+
+"""MIDDLEWARE API
```
claudio525, line 87: I think endpoint functions should be at the top and then make all non-endpoint functions protected (i.e. add _ to the front of their name).
Reason: The main purpose of the file is to define/create the endpoints, hence that code should come first. All other functions are helper functions, and if this is only place where you are planning on using them then they should be protected/private. Otherwise they should probably go into another file.

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+
+# Will be uesd for User dropdown
```
claudio525, line 90: This should be in the function docstring

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+@app.route("/middlewareAPI/auth0/user/get", methods=["GET"])
+@requires_auth
+def get_all_user_from_auth0():
+    """Retrieve all the existing users from the Auth0"""
+    return jsonify(get_users())
+
+
+# Will be used for Project dropdown
```
claudio525, line 98: This also belongs in the docstring

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
+from flask import jsonify, request
```
claudio525, line 1: Order of imports, https://www.python.org/dev/peps/pep-0008/#imports
I don't follow this super strictly, but I do like the ordering as described here:
```
Imports should be grouped in the following order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

You should put a blank line between each group of imports.
```
Additionally in each block I like having the `import x` ones before the `from x import y` ones.

I.e. I would re-write the below as:
```python
import os
import json
from http import client

import requests
from flask import jsonify, request

from ..server import bla, bla, bla
```

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+    get_addable_projects,
+)
+
+import os
+import json
+import http.client
+import requests
+
+# To communicate with Management API
+AUTH0_CLIENT_ID = os.environ["AUTH0_CLIENT_ID"]
```
claudio525, line 15: I think these belong into server.py and then just import it from there

Also given the number of imports from server.py, it might be worth changing it to
`from .. import server`

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+def get_users():
+    """Get all users"""
+    resp = requests.get(
+        AUTH0_AUDIENCE + "users",
+        headers={"Authorization": "Bearer {}".format(get_management_api_token())},
+    )
+
+    # List of dictionaries
+    user_list = resp.json()
+
```
claudio525, line 55: Unnecessary blank line, you can also define both of these variables in one line:
`user_list, user_dict = resp.json(), {}`

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+    user_dict = {}
+
+    # We want to store in a dictionary in form of
+    # { user_id : email | provider}
+    # The reason we keep both email and provider is due to preventing confusion based on having the same emails but different provider
```
claudio525, line 60: This line is probably slightly to long?

middleware/middleware/api/middleware_api.py
```diff
@@ -0,0 +1,116 @@
...
+    # For instance, email A with Google and email A with Auth0
+
```
claudio525, line 62: Don't need a blank line here (since the comment is for the for loop)

middleware/middleware/server.py
```diff
@@ -161,9 +152,33 @@ def get_available_projects():
...
+    return available_projects
+
+
+def get_projects_from_project_api():
+    """Get a list of Project IDs & Project Names from Project API.
+    (Available Projects that we currently have, not from the DB.)
+    Form of
+    {project_id: {name : project_name}}
+    """
+    all_projects_dicts = proxy_to_api(request, "api/project/ids/get", "GET").get_json()
```
claudio525, line 164: Nothing wrong with what you have, but personally I would make this function a one liner, i.e.
`return  proxy_to_api(request, "api/project/ids/get", "GET").get_json()`

middleware/middleware/server.py
```diff
@@ -161,9 +152,33 @@ def get_available_projects():
...
+
+    return all_projects_dicts
+
+
+def get_available_projects():
+    """Do cross-check for the projects.
+
+    It finds available projects from the DB (Available_Project that contains user_id and project_name).
```
claudio525, line 172: I think this line and 174 are both to long (pycharm should highlight these for you?)

middleware/middleware/server.py
```diff
@@ -176,6 +191,99 @@ def get_available_projects():
...
 
 
+def get_addable_projects(query_id):
+    """Similar to the get_available_projects above.
+
+    get_available_projects is there to do the cross-check for the Project tab,
+    compare DB and Project API to see whether users actually have permission to access.
+
+    This function, get_addable_projects is for Edit User feature in the frontend.
+    It compares the projects between the DB and Project API and return the options that are not in the intersction.
```
claudio525, line 201: Line length, I think once its over 80 I usually start wrapping

middleware/middleware/server.py
```diff
@@ -176,6 +191,99 @@ def get_available_projects():
...
+
+
+def check_user_in_db(user_id):
+    """To check whether the given user_id is in the DB"""
+    return bool(User.query.filter_by(user_id=user_id).first())
+
+
+def add_user_to_db(user_id):
+    """Add an user to the MariaDB if not exist"""
+    if check_user_in_db(user_id) == False:
```
claudio525, line 233: Slightly better way to write this would be `if not check_user_in_db(..)` and renaming the function to user_in_db or is_user_in_db. Or if you want to be more explicit use `is` instead of `==` for bool checks.

middleware/middleware/server.py
```diff
@@ -176,6 +191,99 @@ def get_available_projects():
...
+        new_user = User(user_id)
+        db.session.add(new_user)
```
claudio525, line 235: I would just make this a one-liner, i.e. `db.session.add(User(user_id))`, since the variable is only used once (and the definition is nice and short)

middleware/middleware/server.py
```diff
@@ -176,6 +191,99 @@ def get_available_projects():
...
+
+
+def add_available_project_to_db(user_id, project_name):
+    """This is where we insert data to the bridging table, available_projects
+    Unlike any other query, to a bridging table, we need to do the following steps:
+    1. Find an User object by using user_id
+    2. Find a Project object by using project_name
+    3. Append(Allocate, they use Append for a bridging table) the User object to the Project object
+    """
+
```
claudio525, line 265: No need for a blank line here

claudio525 (commented):
Looks good, have a bunch of comment, pretty much all of them are just style. Just thought it's good to point these out now, especially as you start to do more python.
Additionally I think it would be worth for you to check out type hints (https://realpython.com/lessons/type-hinting/) and docstring format. There are lots different docstring formats, I use the numpy one (you can set this up in pycharm to generate a template automatically), see https://numpydoc.readthedocs.io/en/latest/format.html. I don't follow this super strictly, but I do like having some structure in docstrings. Also, if the function is trivial, then I don't usually worry about a docstring and just add a one liner. But for a function with a lot of arguments and some non-trivial functionality I think it is def worth adding a quality docstring. All about finding a balanced approach for this imo.

jasonmotha (dismissed):

middleware/middleware/server.py
```diff
@@ -72,22 +79,68 @@ def __init__(self, error, status_code):
...
 @app.errorhandler(AuthError)
 def handle_auth_error(ex):
     response = jsonify(ex.error)
     response.status_code = ex.status_code
     return response
 
 
-def get_user_id():
-    """We store Auth0 id to DB so no need extra step, just pull sub value which is the unique user_id"""
+def _get_management_api_token():
```
claudio525, line 89: Hmh I don't think these functions belong into server.py either, lets talk Monday!

sungeunbae:
> Additionally I think it would be worth for you to check out type hints (https://realpython.com/lessons/type-hinting/) and docstring format. There are lots different docstring formats, I use the numpy one (you can set this up in pycharm to generate a template automatically), see https://numpydoc.readthedocs.io/en/latest/format.html.

Just out of curiosity - Does VScode have this setup too?

claudio525:
@sungeunbae https://github.com/NilsJPWerner/autoDocstring

middleware/middleware/service.py
```diff
@@ -0,0 +1,421 @@
+import json
```
claudio525, line 1: I suggest renaming this file to db.py, and moving the functions `requires_permission`, `proxy_to_api` and `get_token_auth_header` to again a separate file called utils.py

middleware/middleware/service.py
```diff
@@ -0,0 +1,421 @@
...
+    AUTH0_DOMAIN,
+    db,
+    coreApiBase,
+    projectApiBase,
+    coreApiToken,
+)
+
+from models import *
+
+# Create tables - It only creates when tables don't exist
```
claudio525, line 23: This belongs into server.py
This file should not execute any code, it only defines functions. 
`server.py` on the other hand is where all the setup happens, hence that is where those two lines belong imo.

Usually code files fall into 2 categories,
1) they do something, e.g. a script that is called from the command line (I usually put these under a scripts folder in a python package)
2) module file that contains functions, but doesn't actually do anything when run (These belong in the src folder). And these are the ones that are used by scripts from the package or external of the package)

These are very rough guidelines and its def not as simple as that, but imo its good habit to aim for this. 
`server.py` is one of those weird ones, I kinda think of it as a setup file. If you remind me next time I'm in the office then I can give you a run-down on python package organization (if you want).

seistech internal is probably not a bad example, https://github.com/ucgmsim/seistech/tree/master/seistech_internal/seistech_internal

middleware/middleware/api/project_api.py
```diff
@@ -1,12 +1,13 @@
 from flask import request
 
-from ..server import (
-    app,
+from ..server import app
+from ..service import (
     proxy_to_api,
-    requires_auth,
     get_available_projects,
 )
```
claudio525: unnecessary blank line

middleware/middleware/api/middleware_api.py
```diff
@@ -1,15 +1,17 @@
...
-    app,
+from ..server import app
+
+from ..service import (
     get_users,
-    requires_auth,
     get_token_auth_header,
     get_addable_projects,
     allocate_users_to_projects,
 )
```
claudio525: unnecessary blank line

middleware/middleware/api/middleware_api.py
```diff
@@ -1,15 +1,17 @@
 from flask import jsonify, request
 from jose import jwt
 
-from ..server import (
-    app,
+from ..server import app
+
```
claudio525: unnecessary blank line

middleware/middleware/api/core_api.py
```diff
@@ -1,13 +1,15 @@
...
-from ..server import (
-    app,
+from ..server import app
+
+from ..service import (
     proxy_to_api,
-    requires_auth,
     requires_permission,
     AuthError,
 )
```
claudio525: unnecessary blank line

middleware/middleware/api/core_api.py
```diff
@@ -1,13 +1,15 @@
 from flask import request
 
-from ..server import (
-    app,
+from ..server import app
+
```
claudio525: unnecessary blank line

middleware/middleware/service.py
```diff
@@ -0,0 +1,420 @@
...
+    request,
+    route,
+    methods,
+    endpoint: str = None,
+    content_type: str = "application/json",
+    headers: Dict = None,
+):
+    """Middleware - Handling the communication between Frontend and Core API.
+    We check the request.full_path (e.g., projectAPI/ids/get)
+    If it contains projectAPI, we siwtch APIBase to Project API path.
```
jasonmotha, line 312: switch

middleware/middleware/service.py
```diff
@@ -0,0 +1,420 @@
...
+            401,
+        )
+
+    parts = auth.split()
+
+    if parts[0].lower() != "bearer":
+        raise AuthError(
+            {
+                "code": "invalid_header",
+                "description": "Authorization header must start with" " Bearer",
```
jasonmotha, line 389: Any reason for the split strings?

jasonmotha (changes requested):

middleware/middleware/utils.py
```diff
@@ -0,0 +1,146 @@
...
+        The default media type is application/json
+    headers: object
+        An object that stores some headers.
+    """
+    APIBase = coreApiBase
+
+    if "projectAPI" in request.full_path:
+        APIBase = projectApiBase
+
+    # If endpoint is specified, its the one with uesrs' insteaction, record to DB
```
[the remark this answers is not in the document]
jasonmotha: Sorry for picking up all your spelling mistakes. They just leap out at me when skimming through. 😅

PR #47, opened 2021-02-17
Documentation Framework

SnowNooDLe:
# Documentation Framework

The page can be accessed via the dropdown.

image

Load files from `assets/documents` to render the list(in the left panel) and the markdown(in the right panel) dynamically.

To achieve this, we need to apply three restrictions to a filename.

1. A number comes first to sort the document in the order we want. (This order will decide which title to be displayed first in the list.)
2. Subheader's title comes after which will be used to display a clickable subheader to expand to show more contents that are related to this subheader's title. E.g., If subheader's title is Hazard, by expanding it, it will show all the documents that are related to the Hazard Analysis tab.
3. Finally, the title of the document that will be displayed once the subdirectory gets expanded.

```
1_Hazard-Site-Selection.md
```

- Number 1 - This number can be used to sort the order we want of items in the list(the left panel)
- Hazard - Subheader's title
- Site-Selection - Certain Markdown's title to be displayed in the menu.

image

So the Site Selection for the Hazard comes first as 1 is the lowest number we have for now.

**Hazard** which is the subheader.

**Site Selection**, the title and by clicking the title, the app will render the content that is inside this file on the right side.

image

https://user-images.githubusercontent.com/7849113/108157451-9f0d5580-7147-11eb-9e59-c5eccc18b2e7.mp4

## One thing to consider

Both columns, they don't support the scroll bar. So when there are too many or out of spaces(in smaller screens), the page gets longer and longer as there is no scroll bar but I don't see that those libraries support the scroll bar. Hence, one option is adding something like "Go to top" button at the bottom of the page.


claudio525 (approved):

frontend/src/views/FrameworkDocView.js
```diff
@@ -0,0 +1,197 @@
...
+  const classes = useStyles();
+
+  const [selectedDoc, setSelectedDoc] = useState({
+    header: "",
+    body: "",
+  });
+
+  const [markdownFilesObj, setMarkdownFilesObj] = useState({});
+
+  // Unfortunatley, we cannot create state hook dynamically, so must be hardcoded to control
```
jasonmotha, line 72: Unfortunately

frontend/src/views/FrameworkDocView.js
```diff
@@ -0,0 +1,197 @@
...
+  root: {
+    width: "100%",
+    backgroundColor: theme.palette.background.paper,
+  },
+  nested: {
+    paddingLeft: theme.spacing(4),
+  },
+}));
+
+// Similar to getPrefisNum, instead we get something after number underscore
```
jasonmotha, line 50: prefix

jasonmotha (approved):

claudio525 (approved):

claudio525 (approved):

PR #48, opened 2021-02-22
Fix the Middleware's bug, Circular import

SnowNooDLe:
# Fix the Middleware's bug, Circular import

Changed the structure of our Middleware API to solve the circular import issue.

## Diagram

The way of reading the following diagram, Coloured texts belong to the same coloured file.

For instance, `from auth0 import something(some methods)` and `from server import something` are in a file called `decorator.py`.(Red texts to the file in red.)

image

1. **models.py** is models for MariaDB.
2. **auth0.py** is for handling anything related to Auth0(To do something with Access Token)
3. **decorator.py** is a decorator (Currently checking, require_auth for certain APIs)
4. **db.py** anything to deal with MariaDB(query related, insert, update, remove etc...)
5. **utils.py** now only contains proxy_to_api function that passes the request from the Middleware to the Core/Project API
6. **custom_sqlalchemy.py** is sort of a temporary file due to removing `db.create_all()` from **server.py** so in case we lose the DB, we can run a script, **create_db.py** to create a DB with all user information from Auth0.

## Bring back the record history functionality

By solving this circular import issue, we can bring this functionality back.

image

## TODO

Eventually, could be better to apply the [Flask Application Factories](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/) to set the config better and perhaps, run the application in a different environment easily. (For instance, local setup will run Flask with `debug=True` else `False`(DEV, EA))


jasonmotha (approved):

jasonmotha:
The import in the middle of server.py is less than ideal but not sure how is best to resolve that one.

claudio525:
> The import in the middle of server.py is less than ideal but not sure how is best to resolve that one.

Hmh isn't that removed in this PR?

SnowNooDLe:
> @jasonmotha
> 
> > The import in the middle of server.py is less than ideal but not sure how is best to resolve that one.
> 
> Hmh isn't that removed in this PR?

He accidentally got confused with the one in `create_db.py` which is the temporary file that will deal with creating the DB if somehow we lose the db.

jasonmotha:
Yeah. Clarified in person. Then my only one was to move create_db to a scripts folder.

claudio525 (approved):

PR #49, opened 2021-02-26
Add Permission Dashboards

SnowNooDLe:
# A few more updates on the Intermediate API - 3rd March 2021,

- Restructure the Intermediate API.
  - Change the way we import packages like the SeisTech package does(CoreAPI/ProjectAPI) to keep the conventions.
  - Rename a few functions name to make them simple but meaningful.
  - Also, updated a few docstrings to make them simple but meaningful, also added Parameters and Returns information.
  - Updated a few tables names to avoid any confusion
    - `Auth0_Permission` is the table that contains permission information from Auth0.
    - `Users_Permissions` is the bridging table between `User` and `Auth0_Permission`.
    - `Users_Projects` is the bridging table between `User` and `Project`. (It used to be called, `Available_Projects` but saying `Available_Projects` was not 100% clear. So even it's not the best practice of naming the bridging table with connected tables but I thought this was the best we could find for now.)
  - Some of the endpoint's path is also changed.

## Change the way we import packages

Based on @claudio525 explanation, we use the absolute path? to import packages. (If packages are in the same directory, you don't necessarily have to do it.)
image

## Updated a few tables name to avoid any confusion

- `Auth0_Permission` is the table that contains permission information from Auth0.
- `Users_Permissions` is the bridging table between `User` and `Auth0_Permission`.
- `Users_Projects` is the bridging table between `User` and `Project`. (It used to be called, `Available_Projects` but saying `Available_Projects` was not 100% clear. So even it's not the best practice of naming the bridging table with connected tables but I thought this was the best we could find for now.)

## Rename a few functions name to make them simple but meaningful

Some of the functions are matching their name to the table so we know that it is returning something from a certain table. 
For instance, `get_all_users_projects`, we now have a table called, `Users_Projects`, hence this function will return every data from the `Users_Projects` table. Same to `get_all_users_permissions` as we also have a table called `Users_Permissions`.

image

image

image

### Download

Instead of doing `/coreAPI/hazard_download`, I decided to put them in a subdirectory structure due to matching what we already have. For instance, we have the following paths:
1. `/coreAPI/hazard/get`
2. `/coreAPI/hazard/nz1170p5/get`
3. `/coreAPI/hazard/nz1170p5/soil_class/get`
4. `/coreAPI/hazard/nz1170p5/default/get`
5. `/coreAPI/disagg/get`
6. `/coreAPI/uhs/get`
7. `/coreAPI/uhs/nz1170p5/get`

Hence, I thought it made more sense to have download under hazard if its download data for Hazard Curve.

image

### Intermediate API endpoints

The same approach to what I did in `Download`, based on the feature, split them into the subdirectory.

For instance, AUTH0 related will be under `auth0`,
USERS_PERMISSIONS table, under `users_projects`
USERS_PROJECTS table, under `users_projects`
and so on...

image

# TODO - Before deploying to AWS

Remove the DB and run the `create_db.py` script to create an inial DB due to a change in our UserDB structure.

We used to have five tables,

image

Now we have seven tables,

image

As we cannot pull all users permission via Auth0, we have to store them in UserDB.

# Small update 1 - PrivateRoute.js

Instead of using a path to check the permission, we now pass the permission prop to check. Hence, with one permission `psha-admin` will be able to handle multiple pages, not just a URL with `/psha-admin`.

# Small update 2 - Refactor the way we import modules

We used to have something like this:
Screenshot from 2021-02-26 09-30-03

And now we have this:

image

No downside of doing one above but looked messy and I had to type more if we import more components.

Also, just like we do in Python, I set some conventions for the way we import packages/modules/libraries in a certain order.

```javascript
// React package always comes first
import React, { Fragment, useContext, useState, useEffect } from "react";

// Line break with third party packages/libraries
import { Tabs, Tab } from "react-bootstrap";
import Select from "react-select";
import dompurify from "dompurify";

// Line break with the following files (CONSTANT, Global variables & useAuth0 is a component 
// but a special and will eventually be removed,
// so keep it with CONSTANT and useContext(global variables) for now
import { GlobalContext } from "context";
import * as CONSTANTS from "constants/Constants";
import { useAuth0 } from "components/common/ReactAuth0SPA";

// Line break with reusable components/functions from components or utils directory
import {
  GMSViewerIMDistributions,
  GMSViewerSpectra,
  GMSViewerMwRrupPlot,
  GMSViewerCausalParameters,
} from "components/Hazard/GMS";
import {
  LoadingSpinner,
  DownloadButton,
  GuideMessage,
  ErrorMessage,
} from "components/common";
import { handleErrors, GMSIMLabelConverter } from "utils/Utils";

// Line break with CSS
import "assets/style/GMSViewer.css";
```

# Add Permission Dashboards

### Project Permission Dashboard

image

### Page Access Permission Dashboard

image

image

## Workflow - Page Permission Dashboard

1. When a user is logged in, right after getting permission from Auth0&MiddlewareAPI(Because with our **Role-Based Access Control**, Access token contains the permission but it is encoded. Hence, send this token to our MiddlewareAPI to decode the token to get permission information.), we send another request to the MiddlewareAPI with the following params:
   1. User_id: Auth0's user id (unique)
   2. Permission_list: Permission list from the token which is the trusted source
2. Then the MiddlewareAPI is trying to update the `granted_permission` table which is used for the **Page Permission Dashboard** and have three different scenarios.
   1. If the token has A, B, C permission but the `granted_permission` table doesn't have, then the MiddlewareAPI updates the `granted_permission` table to contain the permission from this user. (Insert data to DB)
   2. If the `granted_permission` table has every permission that token has, do nothing. (Do nothing as no needed)
   3. If the token has A, B, C permission but the `granted_permission` table has the permission of A, B, C, D then the MiddlewareAPI will remove this permission D from the `granted_permission` table as the token is the trusted source of permission. (Remove data from DB)

### Example

User 5f4864800e634f006d1d5710, does not have permission of **create-project**.

Screenshot from 2021-02-26 11-35-45

So it's cell is filled with red colour.

Screenshot from 2021-02-26 11-35-51

Allocate the **create-project** permission from Auth0 and when this user is logged in, the DB gets updated.
(5f4864800e634f006d1d5710 now has permission of **create-project**)

Screenshot from 2021-02-26 11-36-22

Hence, the cell is now green.

Screenshot from 2021-02-26 11-36-43

Now, I removed this user's permission from Auth0 and DB gets updated not to have **create-project** because the trusted source, token tells DB that this user no longer has the permission of **create-project**.

Screenshot from 2021-02-26 11-37-19

Hence, it got back to red.

Screenshot from 2021-02-26 11-37-38

## Project Permission Dashboard

Project Dashboard was easy as we already implemented most features by having the tab, **Edit Users Permission** which updates the `available_project` table. 
And this Project Dashboard pulls data from this `available_project`, so not much fancy work applied on this dashboard.


claudio525 (commented):
Generally looks good, I have some comments regarding naming and the endpoints. But I think it might be much easier if we chat regarding this in person. @SnowNooDLe Are you happy to hold fire on this until Monday and then we can have a chat?

SnowNooDLe:
> Generally looks good, I have some comments regarding naming and the endpoints. But I think it might be much easier if we chat regarding this in person. @SnowNooDLe Are you happy to hold fire on this until Monday and then we can have a chat?

For sure.

frontend/src/components/PermissionConfig/ProjectPermissionDashboard.js
```diff
@@ -0,0 +1,256 @@
...
+  const [userOption, setUserOption] = useState([]);
+
+  const [allProjects, setAllProjects] = useState({});
+  const [tableHeaderData, setTableHeaderData] = useState([]);
+
+  const [allAvailableProjects, setAllAvailableProjects] = useState({});
+  const [tableBodyData, setTableBodyData] = useState([]);
+
+  /*
+    Fetchig All projets we have from Project API to become a table's header
```
jasonmotha, line 24: fetching

jasonmotha (approved):

jasonmotha (commented):
In general looks good. However it is hard to see the new module layout.

middleware/middleware/api/intermediate_api.py
```diff
@@ -0,0 +1,136 @@
...
+@decorator.requires_auth
+def get_all_users_projects():
+    """Get allowed project permissions for all users"""
+    return db.get_all_users_projects()
+
+
+@server.app.route(const.INTERMEDIATE_API_ALL_PERMISSIONS_ENDPOINT, methods=["GET"])
+@decorator.requires_auth
+def get_all_permissions():
+    """Pull every permission from Auth0_Permission table"""
```
claudio525, line 126: `Pull all possible access permission (Auth0_Permission table)`
To make clear that this function doesn't return all permissions for all users, but instead returns all possible permission that can be assigned.

middleware/middleware/auth0.py
```diff
@@ -68,6 +72,57 @@ def get_user_id():
...
+    """Get all users"""
+    resp = requests.get(
+        server.AUTH0_AUDIENCE + "users",
+        headers={"Authorization": "Bearer {}".format(_get_management_api_token())},
+    )
+
+    # List of dictionaries
+    user_list, user_dict = resp.json(), {}
+
+    # We want to store in a dictionary in the form of
```
claudio525, line 109: I think most of this block comment can go in the docstring?

middleware/middleware/create_db.py
```diff
@@ -20,17 +20,35 @@
 app.config["SQLALCHEMY_DATABASE_URI"] = DATABASE
 app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
 
-db = CustomSQLALchemy(app)
+db = cs.CustomSQLALchemy(app)
 
-from models import *
+from middleware.models import *
```
claudio525, line 25: I think this should be changed to `import middleware.models as models`? 
`*` imports are bad (https://stackoverflow.com/questions/2360724/what-exactly-does-import-import)

claudio525 (commented):
Some nitpicking, but this is looking great. Good work applying all the feedback!!

middleware/middleware/__init__.py
```diff
@@ -1,2 +1,2 @@
 from .server import *
```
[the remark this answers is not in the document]
jasonmotha: I think this is what he was talking about yesterday with the DB. I think having a DB class as an interface for a global DB is a good compromise. But let's chat Monday.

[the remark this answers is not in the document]
claudio525: Sweet as, lets chat Friday. Don't worry about changing anything for this then @SnowNooDLe

PR #50, opened 2021-03-09
Mod suggestions

SnowNooDLe:
# Restructure the Middleware

## Make more use of `__init__.py`

By removing `server.py`, `__init__.py` becomes more useful.
Separate the constant variables to files where it belongs. (For instance, anything regarding Auth0 is now sitting in the `auth0.py`)

## Remove `__init__.py` from `/middleware/api/` directory

By seeing the reference document, it seems `__init__.py` file is no longer needed inside the api directory as having `__init__.py` is one of the solutions to let the module knows where those files are(files under /middleware/api) However, we don't actually call any functions(endpoints) by their name, they are called by the URL. Hence, instead of having an empty `__init__.py`, we import files inside another `__init__.py` under /middleware.

[reference](https://timothybramlett.com/How_to_create_a_Python_Package_with___init__py.html)

## proxy_to_api() takes a few more parameters

image

1. We used to call `auth0.get_user_id()` inside this `proxy_to_api()` function which was not the best practice. Hence, from the endpoints level, we create the `user_id` variable, then pass it to lower levels, (to `proxy_to_api()` in this case.)
image
2. As we are planning to have four different tokens (DEV/EA and CORE/PROJECT API), we also pass those parameters to the `proxy_to_api()`.
3. rename `endpoint` param to `action` to avoid any confusion.
4. `api_destination` param to tell this `proxy_to_api` so it can forward the request to either CORE or PROJECT as we expected.

## `db.py` only cares about any interactions with our DB

image

For instance, function called `allocate_projects_to_user()` (In the right image), it inserts data into `Users_Projects` table. Then inside this function, it does not have to return anything as it's just updating the DB.

However, this function gets called by the endpoint, called `allocate_projects_to_user()`(In the left image)
- I know that we don't actually call this function by its name, this gets called by the app.route 

and because this is one of the endpoints, we need to return something (To meet the Flask requirements). Hence, instead of having

`return db.allocate_projects_to_user(user_id, project_list)` inside this function(One in endpoints) with
`allocate_projects_to_user()` to return `Response(status=200)`.

We make the endpoint to return `Response(status=200)` as endpoints are the ones dealing with Flask, not the `db.py`


middleware/middleware/api/core_api.py
```diff
@@ -63,15 +91,18 @@ def get_hazard():
     )
 
 
-@server.app.route(const.CORE_API_HAZARD_NZ11705_ENDPOINT, methods=["GET"])
-@decorator.requires_auth
-def get_hazard_nzcode():
+@app.route(const.CORE_API_HAZARD_NZ11705_ENDPOINT, methods=["GET"])
+@decorators.requires_auth
+def get_hazard_nz_code():
```
jasonmotha, line 96: Do we want to change this nz_code -> nz1170p5?

middleware/middleware/decorators.py
```diff
@@ -1,22 +1,25 @@
...
 
 def requires_auth(f):
     """Determines if the Access Token is valid"""
 
     @wraps(f)
     def decorated(*args, **kwargs):
         token = auth0.get_token_auth_header()
-        jsonurl = urlopen("https://" + server.AUTH0_DOMAIN + "/.well-known/jwks.json")
-        jwks = json.loads(jsonurl.read())
+        json_url = urlopen("https://" + auth0.AUTH0_DOMAIN + "/.well-known/jwks.json")
```
jasonmotha, line 21: Tempted to suggest urllib unparse. But it's probably more clunky than string addition.

middleware/middleware/utils.py
```diff
@@ -36,14 +41,10 @@ def proxy_to_api(
         An object that stores some headers.
     """
 
-    api_destination = server.CORE_API_BASE
-
-    if to_project_api is True:
-        api_destination = server.PROJECT_API_BASE
-
-    if endpoint is not None:
+    if action is not None and user_id is not None:
```
jasonmotha, line 44: Would `if action and user_id` be appropriate?

jasonmotha (changes requested):

middleware/middleware/api/core_api.py
SnowNooDLe: I think to be on the same page, yeah better to have the same word.
Would it be
`nz1170p5`? or `nzs1170p5`?

jasonmotha (approved):

middleware/middleware/api/core_api.py
claudio525: nzs1170p5?

jasonmotha (approved):

claudio525 (approved):

PR #51, opened 2021-03-10
Project Access Level - Part 1 - Refactor the IntermediateAPI & Frontend

SnowNooDLe:
# Project Access Level - Part 1 - Refactor the IntermediateAPI & Frontend

## Update the Project table to have the following attributes:
  1. project_code - E.g., gnzl, mac_raes, nzgs_pga...
  2. project_name - Generic New Zealand Locations, MacRaes Oceania Gold...
  3. access_level - At the moment, only public or private

image

## Update `create_db.py` script

```python
class Project(db.Model):
    project_id = db.Column(db.String(100), primary_key=True)
    project_name = db.Column(db.String(100), nullable=False)
    access_level = db.Column(db.String(100))

    users = db.relationship("UserProject", back_populates="project")

    def __init__(self, project_id, project_name, access_level="private"):
        self.project_id = project_id
        self.project_name = project_name
        self.access_level = access_level
```

Due to restructuring the Project table, updated our `create_db.py` script.

If I recall correctly from yesterday's discussion, when we insert a new Project to the project table, unless it's specified, we are going to set their access_level as PRIVATE. Hence, our Project class's constructor has a default value, `access_level="private"`

image

And We currently have five projects to run(two public, three private). Then depends on the project code, insert the project into the Project table with PUBLIC access_level.

## Frontend Update

**Addable Projects**, **Allowed Projects** and **Project Permission Dashboard** are functioning as expected with the updated Project table. The reason you see NZGS under **Addable Projects** and as red colour in the dashboard is because I haven't touched the logic part yet. We don't care whether projects are public or private at this point, but after this, the logic will be updated then we do care whether they are private or public.

For instance, for **Addable Projects**, the frontend will send a few requests to the Intermediate API to get the following data:
1. User's Projects (query from UsersProjects with user_id)
2. All Projects (query from Project table with all(), we don't care whether they are public or private)

Then the frontend will have some logic to do something like (**Upcoming PR, not the current PR**)
1. Filter the All Projects with their access level, create an array/object with only projects that have private access_level
2. Grab User's projects (1 from above)
3. Compare 1 and 2
4. Return an array of projects that can be allocated to a user, ( 1 - 2)

image

image

image


middleware/middleware/db.py
```diff
@@ -31,33 +31,33 @@ def _add_user_to_db(user_id):
         print(f"User {user_id} already exists")
 
 
-def _is_project_in_db(project_name):
+def _is_project_in_db(project_code):
```
claudio525, line 34: We should be consistent with our naming, so I think this (and all the ones below) should be `project_id`

claudio525 (approved):

sungeunbae (approved):

jasonmotha (approved):

PR #52, opened 2021-03-10
Replace frontend nzcode nzs1170p5

SnowNooDLe:
# Replace NZCode to NZS1170p5

## Hold until PR50 goes through(Mods-suggestion)


jasonmotha (approved):
cool! anyones that we missed this time we can change as we see them.

claudio525 (approved):

PR #53, opened 2021-03-12
Project access level update logic

SnowNooDLe:
# Project access level update logic

## Please wait for this [PR51](https://github.com/ucgmsim/seistech_psha_frontend/pull/51) to go through first.

As we discussed last time, there is some change in the middleware.

image

image

## The endpoints do stay as simple as possible

As we agreed on keeping endpoints as simple as possible, all our endpoints are literally just pulling data from DB
(For instance, if you look at `get_private_projects()` then it literally read the Project table and return all rows with access_level = private)

Then, move the logic from the backend to the frontend.
(For instance, to draw the **Project Permission Dashboard** which is required assigned private projects + all public projects, then to achieve this, we send a few different requests to the middleware:
1. Get all public projects from the Project table
2. Get all private projects from the Project table
3. Get all assigned private projects for all user

We used to have one endpoint that deals within the backend and instead of letting the backend does the job, we do most logic in the frontend. EXCEPT, there will be some exceptions from time to time and we do have an exception for this update

We send one request to get ProjectIDs for the project tab. Project IDs for the Project tab is simply
All assigned private projects + all public projects. However, we need to make sure that those project IDs from the project table are also valid.
(For instance, the project table has the following projects: A, B, C but we noticed one issue from the Project ID = A hence we temporarily remove it from Project API, then the user should not see the Project ID A as it is not valid at the moment. So to make sure that all projects are valid (with some values) and at the same time, we do not want to expose all the valid Project IDs that are sitting in the Project API. Hence, we apply the logic on the middleware to filter out to return some values for the frontend.)

## IntermediateAPI Endpoints

image

## One exception to let the backend has a logic

Where the cross-check happens between the DB and the ProjectAPI data

image


middleware/middleware/api/intermediate_api.py
```diff
@@ -47,34 +45,29 @@ def get_auth0_users():
...
+    """Fetching all public projects from the Project table"""
+    return jsonify(db.get_certain_access_level_projects("public"))
 
 
 @app.route(const.INTERMEDIATE_API_USER_PROJECTS_ENDPOINT, methods=["GET"])
 @decorators.requires_auth
-def get_user_projects():
-    """Fetching all the projects that are already allocated to a user
-    Will be used for Allowed Projects dropdown
+def get_user_assigned_projects_from_db():
```
claudio525, line 64: `get_allowed_projects` or `get_user_allowed_projects`

middleware/middleware/api/project_api.py
```diff
@@ -23,7 +24,7 @@
 
 
 # Site Selection
-def get_all_projects():
+def _get_all_projects():
```
claudio525, line 27: `_get_available_projects`

middleware/middleware/db.py
```diff
@@ -102,7 +102,7 @@ def _is_user_permission_in_db(user_id, permission):
 
 def _add_user_permission_to_db(user_id, permission):
```
claudio525, line 103: `_insert_access_permission`

middleware/middleware/db.py
```diff
@@ -31,33 +31,33 @@ def _add_user_to_db(user_id):
...
+    project_id: string
         A project code we use internally.
         E.g., gnzl, mac_raes, nzgs_pga, soffitel_qtwn...
     """
-    return bool(models.Project.query.filter_by(project_name=project_name).first())
+    return bool(models.Project.query.filter_by(project_id=project_id).first())
 
 
-def _add_project_to_db(project_name):
+def _add_project_to_db(project_id):
```
claudio525, line 46: `_insert_project`

middleware/middleware/db.py
```diff
@@ -128,65 +128,66 @@ def _add_user_permission_to_db(user_id, permission):
         print(f"{user_id} already has a permission with ${permission}")
 
 
-def _add_user_project_to_db(user_id, project_name):
-    """Insert data(allowed projects) to the bridging table,
-    allowed_projects
+def _add_user_project_to_db(user_id, project_id):
```
claudio525, line 131: `_insert_project_permission`

middleware/middleware/db.py
```diff
@@ -128,65 +128,66 @@ def _add_user_permission_to_db(user_id, permission):
...
 
-        db.session.delete(allowed_projects_row)
+        db.session.delete(users_projects_row)
         db.session.commit()
         db.session.flush()
     except:
         print("Something went wrong.")
 
 
 def _remove_user_permission_from_db(user_id, permission):
```
claudio525, line 189: _remove_access_permission

middleware/middleware/db.py
```diff
@@ -198,8 +199,8 @@ def _remove_user_permission_from_db(user_id, permission):
     db.session.flush()
 
 
-def get_user_projects(user_id):
-    """Retrieves all projects ids for the specified user
+def get_user_assigned_projects_from_db(user_id):
```
claudio525, line 202: `get_user_assigned_projects`

middleware/middleware/db.py
```diff
@@ -208,73 +209,41 @@ def get_user_projects(user_id):
...
         .all()
     )
 
-    return [project.project_name for project in allowed_project_objs]
+    return {
+        project.project_id: project.project_name for project in assigned_project_objs
+    }
 
 
 def get_all_users_projects():
```
claudio525, line 226: `get_all_users_project_permissions`

middleware/middleware/db.py
```diff
@@ -459,3 +429,27 @@ def write_request_details(user_id, action, query_dict):
             db.session.add(new_history)
 
     db.session.commit()
+
+
+def get_certain_access_level_projects(access_level):
```
claudio525, line 434: `get_projects` and make access_level optional

middleware/middleware/utils.py
```diff
@@ -76,60 +76,28 @@ def proxy_to_api(
     return response
 
 
-def get_user_projects(db_user_projects, api_projects):
-    """Compute cross-check of allowed projects for the specified user
-    with the available projects from the projectAPI
+def get_user_projects(db_user_projects, public_projects, project_api_projects):
```
claudio525, line 79: `run_project_crosscheck`

claudio525 (dismissed):

frontend/src/components/PermissionConfig/EditUserPermission.js
```diff
@@ -61,7 +61,7 @@ const EditUserPermission = () => {
   useEffect(() => {
     /*
       If fetched addable projects are more than 0, then display them as options.
-      If it is 0, then it is either API is not working or there are no more projects which can be added to this user.
+      If it is 0, then it is either API is not working or there are no more projects which can be removed from this user.
```
jasonmotha, line 64: I think this comment is misleading. Removable projects are in the DB not the API?

frontend/src/components/PermissionConfig/PagePermissionDashboard.js
```diff
@@ -17,11 +17,11 @@ const PagePermissionDashboard = () => {
   const [allPermission, setAllPermission] = useState([]);
   const [tableHeaderData, setTableHeaderData] = useState([]);
 
-  const [allGrantedPermission, setAllGrantedPermission] = useState({});
+  const [allUsersPermission, setAllUsersPermission] = useState({});
   const [tableBodyData, setTableBodyData] = useState([]);
 
   /*
-    Fetchig All projets we have from Project API to become a table's header
+    Fetchig All projets we have from Auth0_Permission table to become a table's header
```
jasonmotha, line 24: Fetching

frontend/src/components/PermissionConfig/PagePermissionDashboard.js
```diff
@@ -146,30 +147,28 @@ const PagePermissionDashboard = () => {
       let tempObj = {};
 
       /*
-      Loop through the object, allGrantedPermission
-      Nested loop through array, allPermission
+      Loop through the object, allUsersPermission
+      Nested loop through the array, allPermission
```
jasonmotha, line 151: I find these kinda comments not that useful as a programmer should be able to infer that from the lines below with the for loop. Like it doesn't give any additional context that the code gives.

middleware/middleware/create_db.py
```diff
@@ -23,6 +23,16 @@
     "psha-admin",
 ]
 
+PROJECT_DICT = {
```
jasonmotha, line 26: Shouldn't this be populated from the API?

jasonmotha (commented):
Thanks for waiting for me.

In general, I don't copy and paste comments. It implies that the documentation is generic and IMO not worth having as comments.

frontend/src/components/PermissionConfig/ProjectPermissionDashboard.js
```diff
@@ -150,41 +173,47 @@ const ProjectPermissionDashboard = () => {
...
+        Nested loop through another object, allProjects
 
         The key for a temp object property with "auth0-user-id", 
         is user-email. (using find function to find an object with auth0-id value
         This object is in the format of
         {value: auth0-id, label: email | auth0 or google auth}
         Then from a found object, use the label as we want email to be displayed
 
-        Then, compare if user's available_projects (projects with permission),
+        Then, compare if user's user_private_projects (projects with permission),
```
sungeunbae, line 185: Minor comment: "Then" appears too often in this block of comment and overall English could be improved a bit.

frontend/src/components/PermissionConfig/EditUserPermission.js
SnowNooDLe: I should've specified more instead of API.

As you said yes, it's from the DB hence if there are no more projects to be added/removed (As we have Add/Remove feature) OR Intermediate API is broken and not functioning properly.

jasonmotha (approved):

frontend/src/components/PermissionConfig/EditUserPermission.js
```diff
@@ -49,7 +49,8 @@ const EditUserPermission = () => {
   useEffect(() => {
     /*
       If fetched addable projects are more than 0, then display them as options.
-      If it is 0, then it is either API is not working or there are no more projects which can be added to this user.
+      If it is 0, then it is either the Intermediate API is not working 
+      or there are no more projects which can be added to this user.
```
claudio525, line 53: A block comment's length should be in relation to the complexity of the code in the code block. Given that this block of code is pretty trivial, the comment should reflect this. I think something like
`Display fetched addable projects` would be enough. Its very easy to see from the code that it defaults to an empty list, so no real reason to highlight that.

Also
>If it is 0, then it is either the Intermediate API is not working or there are no more projects which can be added to this user.


this is not related to the code, this is bigger picture and doesn't really belong in a block comment.

Note: The above is just my opinion, there is no "right" and "wrong" in these sort of things, but imo its all about finding a balance. And usually its that people don't comment enough imo, with you being the exception (you should take that as a compliment haha).

claudio525: Hopefully the above makes sense, happy to clarify if needed.

frontend/src/components/PermissionConfig/EditUserPermission.js
```diff
@@ -60,8 +61,9 @@ const EditUserPermission = () => {
 
   useEffect(() => {
     /*
-      If fetched addable projects are more than 0, then display them as options.
-      If it is 0, then it is either API is not working or there are no more projects which can be removed from this user.
+      If fetched removable projects are more than 0, then display them as options.
```
claudio525, line 64: Same sentiments as above

frontend/src/components/PermissionConfig/PagePermissionDashboard.js
```diff
@@ -21,7 +21,7 @@ const PagePermissionDashboard = () => {
   const [tableBodyData, setTableBodyData] = useState([]);
 
   /*
-    Fetchig All projets we have from Auth0_Permission table to become a table's header
+    Fetching All projets we have from Auth0_Permission table to become a table's header
```
claudio525, line 24: I think if its just a single line comment, just use `// Fetching ...` to make it a comment (saves two lines)

At least I hope thats a thing in javascript

claudio525 (commented):
To further explain my first comment regarding comments. Comments are great, and I swear by them. In my opinion the aim of a comment is (in most cases) not to explain the code that follows, but instead make code navigation easy (its much easier to read a one-liner comment then it is to read 10 lines of code (even it if its trivial code)). If someone wants to know the details of a block of code, they will have to look at the code in details anyways. But lots of comments make it very easy to find the correct block of code that one wants to debug/modify etc.
Anyways, just my 5 cents.

frontend/src/components/PermissionConfig/PagePermissionDashboard.js
SnowNooDLe: By looking at what I've done, seems if its for functions or `useEffect` (Which is a React hook), I put /* */ to explain regardless of lines of codes, like we use """ """ docstring for functions/classes in Python // for comments inside them.

So it looked liked I just used multiline(block comments) /* */ and single-line comment //, inside them.

But I am happy to keep the what JS does, ->
if the comment is only one line, use `//`
if the comment is longer than one line, user `/* */`

frontend/src/components/PermissionConfig/EditUserPermission.js
```diff
@@ -196,9 +185,7 @@ const EditUserPermission = () => {
     };
   }, [selectedUser]);
 
-  /*
-    Allocating selected projectst to the chosen user
-  */
+  // Allocating selected projectst to the chosen user
```
jasonmotha, line 188: projects

frontend/src/components/PermissionConfig/PagePermissionDashboard.js
```diff
@@ -20,9 +20,7 @@ const PagePermissionDashboard = () => {
   const [allUsersPermission, setAllUsersPermission] = useState({});
   const [tableBodyData, setTableBodyData] = useState([]);
 
-  /*
-    Fetching All projets we have from Auth0_Permission table to become a table's header
-  */
+  // Fetching All projets we have from Auth0_Permission table to become a table's header
```
jasonmotha, line 23: all projects

jasonmotha (approved):

claudio525 (approved):

PR #54, opened 2021-03-15
Update GMS in Hazard Analysis tab

SnowNooDLe:
# Update GMS in Hazard Analysis tab

## March 24th updates
1. Instead of performing a deep copy of a whole object, create a new array(Hence, can avoid by reference issue) for a certain value that gets modified inside the function.
In this case, we don't have to spend extra time&memory to perform a deep copy.

```javascript
const periods = [...gmsData["IMs"]];
```
2. Tried to tidy up some comments and instead of explaining what code does, more of giving guidance.

## Fix the GMS plot - instead of using the user's IM/Exceedance rate level, use the IM_j value from the API.

GMS used to plot with users input when the `IM Type(Conditioning IM Name)` starts with `pSA`. Instead of using the user's input, we used the `im_j` value from the response data(From the Core API).

### Example - Previous version

image

### Example - Current version

image

## Remove the `Highlight.js` package as no longer needed.

It used to be used on the `Profile` page. However, this was a default thing from Auth0 and we will have our version of the `Profile` page later.

## Implement the computed GMS data validation
  - Check the IM Type matches with selected Conditioning IM Name
  - Checking its properties
    - To see if any of the properties with 0 lengths (It means no values for this property)
    - To see if some of the properties' properties are matching with the selected IM Vectors
      (For instance, if selected IM Vectors are PGA, pSA_0.01 and pSA_0.02, then properties of `gcim_cdf_x`, `gcim_cdf_y` and `realisations` should match with the selected IM Vectors))
- With the validation above, if something goes wrong, we set the `setShowErrorMessage` to show an error message.
  - Currently, this is more developers-friendly

```javascript
/*
    Lastly, check whether this object has the properties we are looking for
    E.g., gmsData object must-have properties of:
    1. gcim_cdf_x
    2. gcim_cdf_y
    3. realisations
    4. selected_GMs
    5. ks_bounds - Checked by the second checker
    6. metadata
    7. im_j - Checked by the second checker
    8. IMs
    9. IM_j - Checked by the second checker
    1 ~ 3 are objects and these should also have
    properties of selected IMVectors
    For instance, selected IMVectors are PGA, pSA_0.01 and pSA_0.03
    Then 1~3 objects must have properties of PGA, pSA_0.01 and pSA_0.03
  */
const validateComputedGMS = (providedGMSData) => {
    // First, check whether we actually got computed GMS (providedGMSData)
    if (providedGMSData === undefined || providedGMSData === null) {
      return false;
    }

    // Second, check each values of the object, providedGMSData
    Object.values(providedGMSData).forEach((x) => {
      // Like ks_bound value is Number and is not working with Object.keys(x).length
      if (!isNaN(x)) {
        return;
      }
      // When an object doesn't have any values
      if (Object.keys(x).length === 0) {
        return false;
      }
    });

    // Compare IMVectors and selected Vectors to see if they are matching
    const sortedSelectedIMVector = GMSIMVector.map((im) => im.value).sort();

    if (!arrayEquals(providedGMSData["IMs"].sort(), sortedSelectedIMVector)) {
      setShowErrorMessage({ isError: true, errorCode: "gms_im" });
      return false;
    }

    // Compare if gcim_cdf_x has the keys of IMs
    if (
      !arrayEquals(
        Object.keys(providedGMSData["gcim_cdf_x"]).sort(),
        sortedSelectedIMVector
      )
    ) {
      setShowErrorMessage({ isError: true, errorCode: "gms_gcim_cdf_x" });
      return false;
    }

    // Compare if gcim_cdf_y has the keys of IMs
    if (
      !arrayEquals(
        Object.keys(providedGMSData["gcim_cdf_y"]).sort(),
        sortedSelectedIMVector
      )
    ) {
      setShowErrorMessage({ isError: true, errorCode: "gms_gcim_cdf_y" });
      return false;
    }

    // Compare if realisations has the keys of IMs
    if (
      !arrayEquals(
        Object.keys(providedGMSData["realisations"]).sort(),
        sortedSelectedIMVector
      )
    ) {
      setShowErrorMessage({ isError: true, errorCode: "gms_realisations" });
      return false;
    }

    // Compare if IM_j equals to the selected IM Type
    if (providedGMSData["IM_j"] !== GMSIMType) {
      setShowErrorMessage({ isError: true, errorCode: "gms_IM_j" });
      return false;
    }

    // Compare if metadata has the keys of mag, rrup, sf and vs30
    if (
      !arrayEquals(
        Object.keys(providedGMSData["metadata"]).sort(),
        Object.keys(CONSTANTS.GMS_LABELS).sort()
      )
    ) {
      setShowErrorMessage({ isError: true, errorCode: "gms_metadata" });
      return false;
    }

    return true;
  };
```

### Example of error if returned data is not valid

This message can definitely be improved at the point where it's close to the final, stable, production version but for debugging purpose but still can handle error to notify us what went wrong, I think this is enough.

image

## Others - A few updates

1. Order of file's content
    1. React Hook comes first
       - For instance, `useState` and `useEffect`
    2. Any functions
2. Try to avoid declaring variables inside the component(not inside the function)
   - Because React's functional components get re-rendered every time the `useState` gets changed, then by using either `let` or `const` will constantly get called hence, instead of using them, just use `useState` because this hook gets only called once at INITIAL render.
   For instance, instead of having
   ```javascript
     let causalParamBounds = {
      mag: {
        min: GMSMwMin,
        max: GMSMwMax,
      },
      rrup: {
        min: GMSRrupMin,
        max: GMSRrupMax,
      },
      vs30: {
        min: GMSVS30Min,
        max: GMSVS30Max,
        vs30: vs30,
      },
    };

    let filterParamsObj = {
      mag_low: Number(GMSMwMin),
      mag_high: Number(GMSMwMax),
      rrup_low: Number(GMSRrupMin),
      rrup_high: Number(GMSRrupMax),
      vs30_low: Number(GMSVS30Min),
      vs30_high: Number(GMSVS30Max),
    };
   ```
   Now we have this
   ```javascript
     const filterParamsObj = {
        mag_low: Number(GMSMwMin),
        mag_high: Number(GMSMwMax),
        rrup_low: Number(GMSRrupMin),
        rrup_high: Number(GMSRrupMax),
        vs30_low: Number(GMSVS30Min),
        vs30_high: Number(GMSVS30Max),
      };

      // Min/Max values for one of the plots
      // Also only gets called within fetching call(Compute GMS) so doesn't matter with number of render gets done
      setCausalParamBounds({
        mag: {
          min: GMSMwMin,
          max: GMSMwMax,
        },
        rrup: {
          min: GMSRrupMin,
          max: GMSRrupMax,
        },
        vs30: {
          min: GMSVS30Min,
          max: GMSVS30Max,
          vs30: vs30,
        },
      });
   ```
3. Each plot now deals with different CLONED objects
Because JavaScript's Object and Array are by REFERENCE not by VALUE.
Hence by passing/using the object to other functions/variables and if any change made, will affect the original object.
Hence, to prevent any issue, cloned the object(deep copy) for each plot:
    1. IM Distribution
    2. Causal Parameter
    3. MwRrup Plot
    4. Spectra
 
4. Spectra calculation is done somewhere else.
As with re-rendering for functional components, every state gets changed, (even by clicking the dropdown within the `IM Distributions` and `Causal Parameters` tab under the viewer, will make this component re-render because it changes the state.)
The calculation isn't big big but is bigger than any other plots and it certainly takes some time to:
   1. Calculate based on given computed GMS Data
   2. Draw with Plotly.js based on this calculated data 
Hence, instead of calculating within the viewer part, used a function to calculate and store the calculated data into `state` Hence, the component doesn't have to re-calculate every time it gets re-rendered because `useState` only called at the INITIAL render and whenever the `setter` gets called which is inside the fetching call(Computing GMS Data) so unless sending another request to fetch GMS data, this state will never get called again. Hence, the app only needs to deal with drawing plots with `Plotly.js` which we cannot stop doing it.

5. `react-error-overlay` package

We can disable the following error overlay which is only visible on the development but our DEV and EA are running under development. So I thought it would be good to implement this package to disable the error message from the EA version.

image

```javascript
if (process.env.REACT_APP_ENV === "EA") {
  stopReportingRuntimeErrors(); // disables error overlays
}
```


frontend/src/constants/Constants.js
```diff
@@ -258,11 +247,53 @@ export const ERROR_SET_DIFF_CODE = {
     ERROR_MSG_TITLE: "Our server is currently having issues.",
     ERROR_MSG_BODY: "Please try again later.",
   },
+
+  // GMS Validation error
+  gms_im: {
+    ERROR_MSG_HEADER: "Ground Motion Selection - IMs",
+    ERROR_MSG_TITLE:
+      "Issue found in the retruend data that properties do not match with selected IM Vector",
```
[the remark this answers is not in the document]
jasonmotha: THere are a few instances with that spelling.

frontend/src/utils/Utils.js
```diff
@@ -53,25 +51,18 @@ export const getPlotData = (data) => {
...
+// Implement x sig figs for numeric float values
 export const renderSigfigs = (fullprecision, sigfigs) => {
   return Number.parseFloat(fullprecision).toPrecision(sigfigs);
 };
 
-/*
-  Create an Array from start to stop by step
-*/
-
+// Create an Array from start to stop by step
```
jasonmotha, line 59: Again, giving a bit more context with this type of comment. Currently it's basically reiterating the line of code.

frontend/src/utils/Utils.js
```diff
@@ -53,25 +51,18 @@ export const getPlotData = (data) => {
...
 export const range = (start, stop, step = 1) =>
   Array(Math.ceil((stop - start) / step) + 1)
     .fill(start)
     .map((x, y) => x + y * step);
 
-/*
-  Create an special aray for react-select
-*/
+// Create an special aray for react-select
```
jasonmotha, line 65: array - Why do we want a special array.

frontend/src/utils/calculations/CalculateGMSSpectra.js
```diff
@@ -0,0 +1,205 @@
...
+
+    // If IM Type starts with pSA, add it to periods
+    if (im_type.startsWith("pSA")) {
+      periods.push(im_type);
+    }
+
+    // Create an object key = IM, value = Period
+    const localPeriods = {};
+    sortIMs(periods).forEach((IM) => {
+      // With out current format, IM starts with pSA, they have period.
```
jasonmotha, line 23: our*
Can also shrink this comment by removing "we" and other conversational english.

frontend/src/utils/calculations/CalculateGMSSpectra.js
```diff
@@ -0,0 +1,205 @@
...
+
+    /*
+      Loop through the period object (That is in the right order).
+      Create another objects in the right order for the following data:
+      - CDF X
+      - CDF Y
+      - Realisations
+      - Selected GMs
+      as response data may not be in the right order.
+      So we can plot them in the right order.
```
jasonmotha, line 45: Another example of a good comment. But can also shrink the last two lines (you mention the right order 3 times)

frontend/src/utils/calculations/CalculateGMSSpectra.js
```diff
@@ -0,0 +1,205 @@
...
+    */
+    for (const [IM, values] of Object.entries(localPeriods)) {
+      // If IM (IM from IM Vectors) is not same as IM_Type, we do different approach.
```
jasonmotha, line 48: Better off merging the comments in the block with this comment. What is your different approach?

frontend/src/utils/calculations/CalculateGMSSpectra.js
jasonmotha: Because I skim through comments, bullet points work really well.

frontend/src/utils/calculations/CalculateGMSSpectra.js
```diff
@@ -0,0 +1,205 @@
...
+        mode: "lines",
+        name: "Realisations",
+        line: { color: "black" },
+        type: "scatter",
+        showlegend: i === 0 ? true : false,
+      });
+    }
+    /*
+      Selected GMs calculation
+      Same as Realisations caluclation above.
```
jasonmotha, line 178: calculation

jasonmotha (changes requested):

frontend/src/components/Hazard/GMS/GMSForm.js
```diff
@@ -45,12 +45,60 @@ const GMSForm = () => {
...
+
+  const [selectedIMType, setSelectedIMType] = useState("");
+
+  const [localNumGMS, setLocalNumGMS] = useState("");
+  const [localWeights, setLocalWeights] = useState({});
+  const [localDatabase, setLocalDatabase] = useState(null);
+  const [localReplicates, setLocalReplicates] = useState(1);
+
+  // DEFAULT FOR DEV WITH pSA_0.03 & 0.08 Exceedance Rate
+  const [localMwMin, setLocalMwMin] = useState(CONSTANTS.DEFAULT_MW_MIN);
```
claudio525, line 69: Huh any specific reason for hardcoding these defaults? Did Brendon ask for this?

frontend/src/components/Hazard/GMS/GMSForm.js
```diff
@@ -281,13 +251,11 @@ const GMSForm = () => {
       if (localIMVector.length !== 0 && getIMWeightsClick !== null) {
         let queryString = `?IM_j=${selectedIMType}&IMs=`;
 
-        // Create a new array from an object.
-        // As localIMVector is an object with properties of label and value
+        // Create a new array from an object to contain only values
         const newIMVector = Array.from(localIMVector, (x) => x.value);
```
claudio525, line 255: This line & the two below, look to me like one block of code (create the array,  add the IMs, remove last comma), i.e. the block creates the request url (from my understanding). In which case the empty lines between these lines should be removed, to highlight that these belong together?

claudio525 (commented):
Looks good!

frontend/src/components/Hazard/GMS/GMSForm.js
SnowNooDLe: Correct, they are all doing something for one purpose, adding the right values to the query string.

frontend/src/components/Hazard/GMS/GMSViewerCausalParameters.js
```diff
@@ -24,10 +24,9 @@ const GMSViewerCausalParameters = ({
     })
     .flatMap((x) => Array(2).fill(x));
 
-  // Create an array with range 0 and 1 with third argument as step
   const rangeY = range(0, 1, 1 / gmsData["metadata"][metadata].length);
 
-  // Then we duplicate each element that is not the first and last index's element
+  // Duplicate the elements except the first and the last element
```
jasonmotha, line 29: Nice!

frontend/src/components/Hazard/GMS/GMSViewerCausalParameters.js
```diff
@@ -53,7 +52,8 @@ const GMSViewerCausalParameters = ({
     },
   ];
 
-  // Add bounds (Min / Max) if selected metadata is not sf
+  // Add bounds (Min / Max)
```
jasonmotha, line 55: I would still put these on the same line as they are related.

jasonmotha (changes requested):

PR #56, opened 2021-03-22
Implement automation of project creation

SnowNooDLe:
# Implement automation of project creation

### ProjectID Checker - [comment](https://github.com/ucgmsim/seistech_psha_frontend/pull/56#discussion_r625328974)

We will use Project Name(Readable/Display name) as a Project ID if Project ID is not present.
Check whether Project ID is None or empty string, if so, use `Name` as an `ID`.
```python
if data.get("id") is None or not len(data["id"].strip()):
```

### Headers - [comment - 1](https://github.com/ucgmsim/seistech_psha_frontend/pull/56#discussion_r625333892) [comment - 2](https://github.com/ucgmsim/seistech_psha_frontend/pull/56#discussion_r625338835)

`headers` prop no longer needed as the only place used was `download_data` where we tried to set the filename. However, the frontend is the one that decides the file name, not the intermediate API. Hence, decided to remove `param` that is no longer needed.

## Frontend inputs

The frontend collects the following information from a user:
1. Display Name
            - Normal input
            - String 
            - Required
2. Project ID (Internal Name, no spaces) 
            - Normal input
            - Can be left blank 
3. Access Level - Optional but default to Private
            - Private
            - Public 
4. Locations 
            - Normal input for Name
            - Number(Float) input for Lat, Lon, VS30
            - Name, Lat, Lon, VS30(multiple VS30 is possible)
            - Locations can be multiple rows
            - Required
5. Package 
            - Radio button
            - PGA
            - PGA + pSA
            - Custom
            - Required 
  
If **Custom** was chosen, then extra inputs required:
1. IMs
            - Dropdown
            - Multi
2. Disagg RPs
            - String - However only numbers with a comma
            - Multi
3. UHS RPs
            - String - However only numbers with a comma
            - Multi

## Middleware - IntermediateAPI

1. If the **project_id** is not provided, the middleware will create a **project_id** from the given **display_name**, make string to lowercase and replace all space( ) to underscore(_).
   - For instance, if the `diplsay_name` was **Hello World** and no `project_id` provided, then the `project_id` would be **hello_world**
2. Check whether this **project_id** is already taken, if so, send back with the status code `409` which presents a conflict. If it is not taken, forward a request to the ProjectAPI.
3. `response_code` is the result of the ProjectAPI, it can tell whether it's broken, something went wrong or created successfully.
4. Based on this `response_code`, the frontend will show different modals to a user.

```python
def create_project():
    """Create new project(s)"""
    data = json.loads(request.data.decode())

    # Use Display Name as Project ID
    # If Project ID is not provided or spaces
    if not data["id"] or data["id"].isspace():
        data["id"] = data["name"].lower().replace(" ", "_")

    if db.is_project_in_db(data["id"]):
        # Code if resource already exists
        return jsonify({"project_cretaed": False, "reason": "Already exists"}), 409
    else:
        response_code = utils.proxy_to_api(
            request,
            const.PROJECT_CREATE_NEW_ENDPOINT,
            "POST",
            PROJECT_API_BASE,
            PROJECT_API_TOKEN,
            data=json.dumps(data),
            user_id=auth0.get_user_id(),
            action="Project Creation - Trying to create project(s)",
        ).status_code

        if response_code == 200:
            db.add_project_to_db(data["id"], data["name"], data["access_level"])
            return jsonify({"project_cretaed": True}), 200
        else:
            return (
                jsonify({"project_cretaed": False}),
                response_code,
            )
```

## Custom section

As far as I am aware of, custom section is currently disabled.
image

#### Successful

image

#### Failed with taken Project ID

image


middleware/middleware/api/intermediate_api.py
```diff
@@ -117,3 +117,39 @@ def get_all_permissions():
...
+    else:
+        response_code = utils.proxy_to_api(
+            request,
+            const.PROJECT_CREATE_NEW_ENDPOINT,
+            "POST",
+            PROJECT_API_BASE,
+            PROJECT_API_TOKEN,
+            data=json.dumps(data),
+            user_id=auth0.get_user_id(),
+            action="Project Creation - Trying to create project(s)",
```
claudio525, line 145: I think something like
`Project Creation - Triggering project generation` is slightly more clearer

middleware/middleware/utils.py
```diff
@@ -34,10 +33,11 @@ def proxy_to_api(
     methods,
     api_destination: str,
     api_token: str,
+    data: dict = None,
     user_id: str = None,
     action: str = None,
     content_type: str = "application/json",
-    headers: Dict = None,
+    headers: dict = None,
```
claudio525, line 40: This was correct, should be using `Dict` and then ideally even specify the type of the key and value of the dictionary, 
e.g. `Dict[str, int]`, means a dictionary with keys of type string and values of type int

middleware/middleware/utils.py
```diff
@@ -52,15 +52,17 @@ def proxy_to_api(
...
         Determining the user
     action: string
         To find out what user is performing
     content_type: string
         Entry-header field indicates the media type of the entity-body sent to the recipient.
         The default media type is application/json
-    headers: object
-        An object that stores some headers.
+    headers: dictionary
+        A dictionary that stores some headers.
```
claudio525, line 65: I suggest avoiding non-descriptive words like `some` they make it confusing to read the code. E.g. in this scenario, after reading the docstring I don't really know what headers are supported and what this argument does. 

At first glance I assumed these were optional headers that were attached to the coreAPI request, but after looking at the code it seems like these headers are attached to the response from the intermediateAPI. So I think it might be a good idea to rename this variable to `response_headers`, and docstring something like
`Headers to attach to the response of the intermediateAPI`

claudio525 (commented):
Looks good in general :+1, just noticed a couple of minor things. And some suggestions regarding comments/docstrings. Might be worth for us to have a quick chat this morning.

Good job, getting this working though!

middleware/middleware/api/intermediate_api.py
```diff
@@ -126,13 +126,12 @@ def create_project():
     data = json.loads(request.data.decode())
 
     # Use Display Name as Project ID
-    # If Project ID is not provided or spaces
-    if not data["id"] or data["id"].isspace():
+    # If Project ID is not provided
+    if data.get("id") is None or not len(data["id"].strip()):
```
jasonmotha, line 130: len == 0?
That might be clearer to read?

jasonmotha (approved):

middleware/middleware/api/intermediate_api.py
SnowNooDLe: Oh no, I was thinking so hard about between `not` and `== 0` and ended up not after I saw it still prints `False` or `True` haha.

I will replace to `== 0` in different branch I am currently working on, simple adding `NZCode` data into the metadata section.(Just adding Z Factor and Soil Class for Projects)

claudio525 (approved):

joelridden (approved):

PR #58, opened 2021-03-25
Creating the current copy of the DEV to EA

SnowNooDLe:


.gitignore
```diff
@@ -0,0 +1,24 @@
...
+
+.env
+.env.dev
+.env.test
+
+# Pthon virtual environment
+/venv
+
+# vscode setting file
+/.vscode
```
jasonmotha, line 20: Some of these can be excluded from your .git/exclude folder rather than the .gitignore file

jasonmotha (changes requested):
Is there any way to autogenerate the yarn / package-lock files? It seems bulky to have 20k lines like this in git.

.gitignore
claudio525: The `exclude` file in `.git/info` folder, is a private .gitignore (i.e. doesn't get commited)

claudio525 (commented):
Agree with @jasonmotha, don't think `package-lock.json` needs to be commited. From my understanding that is generated when one runs `npm install`, which only needs `package.json`?

jasonmotha (approved):

claudio525 (approved):

PR #61, opened 2021-03-31
Project GMS

SnowNooDLe:
# Projects/Non-projects - GMS - 31st March Update

- Project GMS input modification - Ground Motions Selection ID is split into two parts
  1. Conditioning IM Name
  2. Exceedanec rate level
- Bring Download data functionality back
  - Can also record a user's inputs for Download Data
  - Updated Hazard Curve & Disaggregation's filename based on the feedback from the first tester. 
- Change the way the middleware cross-check to get available projects for a user.
- Tidy up comments/code(unnecessary or duplicates)

# Projects tab - GMS - 26th March Update

Projects GMS is implemented, Tooltips need to be updated also the disclaimer? at some stage.

image

image

image

image

image

image

## .gitignore update - requirements.txt

As we recently changed the workflow of the branches, master_dev and master_ea

master_dev treats as a playground, when a feature seems working, just merge it into master_dev and deploy it to the psha-dev.

Then once it looks good and ready for the review, open PR to master_ea and get a code review.

Hence, the psha-dev will be based on the `master_dev` and the psha-ea will be based on the `master_ea`. Hence the middleware package will be installed from a different branch depends on the deployed branch:
- psha-dev's middleware will be installed from master_dev
- psha-ea's middleware will be installed from master_ea


jasonmotha (dismissed):

frontend/src/components/Hazard/SeismicHazard/UHSSection.js
```diff
@@ -11,7 +11,7 @@ import { GuideTooltip } from "components/common";
 import {
   renderSigfigs,
   disableScrollOnNumInput,
-  checkIMwithPSA,
+  IMStartsWithPSA,
```
claudio525, line 14: I think this function should be called `is_pSA` or `isPSA` (depending on your naming scheme)

A note in general, don't name the function after what the code does (i.e. checking if IM starts with pSA), but instead name it with respect to the goal of the function or overall functionality (i.e. checking if the IM is pSA (goal), which is done by checking if the IM starts with pSA (implementation)).

To put it in a slightly different way, the function name should say what the function does (check if the IM is pSA), but not not how it is done (by checking if it starts with pSA).

Probably worth discussing this briefly. @jasonmotha thoughts?

claudio525 (commented):
One minor comment, otherwise looks good :+1:

claudio525 (dismissed):

frontend/src/components/Project/SeismicHazard/HazardViewerUHS.js
```diff
@@ -186,12 +180,7 @@ const HazardViewerUHS = () => {
         downloadToken={{
           uhs_token: downloadToken,
         }}
-        extraParams={{
-          project_id: projectId,
-          station_id: `${projectLocationCode[projectLocation]}_${projectVS30}`,
-          rp: getSelectedRP().join(","),
-        }}
-        fileName="uniform_hazard_spectrum.zip"
+        fileName="Uniform_Hazard_Spectrum.zip"
```
jasonmotha, line 183: @claudio525 Is it a spectra or a spectrum? What's the difference?

jasonmotha (dismissed):

frontend/src/components/Project/SeismicHazard/HazardViewerUHS.js
claudio525: To me, they are the same thing. But there might be a difference

claudio525 (approved):

PR #69, opened 2021-04-19
Projects' GMS - Mag/Rrup disaggregation distribution plots

SnowNooDLe:
# Projects' GMS - Mag/Rrup disaggregation distribution plots

### Please review the [PR#68](https://github.com/ucgmsim/seistech_psha_frontend/pull/68) first

For **Projects - GMS**, two more plots are added to the **IM Distributions** tab.
    - Magnitude disaggregation distribution
    - Rupture distance disaggregation distribution

image
image


middleware/middleware/utils.py
```diff
@@ -153,3 +154,28 @@ def run_project_crosscheck(db_user_projects, public_projects, project_api_projec
...
+    
+    Returns
+    -------
+    Two sorted arrays
+    """
+    sorted_array = np.rec.fromarrays([np.array(distribution), np.array(contribution)])
+    sorted_array.sort()
+
+    # Use np.cumsum to get cumulative probability(CDF)
+    sorted_distribution, sorted_contribution = (
```
jasonmotha, line 176: split the assigment into two lines?

(claudio will probably just say return them as below but I like the implicit commenting)

jasonmotha (approved):

jasonmotha (approved):

frontend/src/components/common/GMS/GMSCausalParamPlot.js
```diff
@@ -3,31 +3,28 @@ import React from "react";
...
     Create a new array to avoid affecting the original object from any modification
-    then sort metadata & duplicate each element
+    then sort metadata & duplicate every element
   */
-  const xRange = [...gmsData["selected_gms_metadata"][metadata]]
-    .sort((a, b) => {
-      return a - b;
-    })
-    .flatMap((x) => Array(2).fill(x));
+  const xRange = sortDuplicateXRange([
```
claudio525, line 15: How come we are duplicating?

middleware/middleware/api/project_api.py
```diff
@@ -182,6 +183,40 @@ def get_gms_default_causal_params():
         action="Project - GMS Get Default Causal Params",
     )
 
+    if response.status_code is 200:
+        response_data = response.get_json()
+        # Keep every key except "contribution_df" key
+        new_dict = dict(
```
claudio525, line 189: You can delete a dictionary entry using: `del d[key`, but you don't actually need to, just overwrite it.
I.e. I would suggest doing the below as follows:
```python
        contribution_df_data = response_data["contribution_df"]

        sorted_rrup, rrup_cdf = utils.calc_cdf(
            contribution_df_data["contribution"][:], contribution_df_data["rrup"]
        )
        sorted_mag, mag_cdf = utils.calc_cdf(
            contribution_df_data["contribution"][:], contribution_df_data["magnitude"]
        )
        response_data["contribution_df"] = {
            "magnitude": sorted_mag.tolist(),
            "mag_contribution": mag_cdf.tolist(),
            "rrup": sorted_rrup.tolist(),
            "rrup_contribution": rrup_cdf.tolist(),
        }
```
In which case I would also suggest renaming `response_data` to something like `data_dict`

middleware/middleware/api/project_api.py
```diff
@@ -182,6 +183,40 @@ def get_gms_default_causal_params():
...
+            "magnitude": sorted_mag.tolist(),
+            "mag_contribution": mag_cdf.tolist(),
+            "rrup": sorted_rrup.tolist(),
+            "rrup_contribution": rrup_cdf.tolist(),
+        }
+
+    else:
+        return Response(status=response.status_code)
+
+    return jsonify(new_dict)
```
claudio525, line 218: This should be in the if, in which case you also don't need the else, i.e.
```python
... 
    return jsonify(...)
return Response(...)
```

middleware/middleware/utils.py
```diff
@@ -153,3 +154,26 @@ def run_project_crosscheck(db_user_projects, public_projects, project_api_projec
...
+    Parameters
+    ----------
+    contribution: array
+    distribution: array
+    
+    Returns
+    -------
+    Two sorted arrays
+    """
+    sorted_array = np.rec.fromarrays([np.array(distribution), np.array(contribution)])
```
claudio525, line 172: Not sure I like this, its not super obvious what is happening I think.
I think argsort is cleaner (and doesn't any array manipulation beforehand) 
i.e.
```python
sort_ind = np.argsort(distribution)
distribution, contribution = distribution[sort_ind], contribution[sort_ind]

return distribution, np.cumsum(contribution)
```
Also I think you should rename `distribution` to `x_values` or `data_values` and `contribution` to `weights`

claudio525 (approved):

claudio525 (commented):
The previous one was meant to be a comment, not approve...

frontend/src/components/common/GMS/GMSCausalParamPlot.js
SnowNooDLe: image
To achieve this sort of stair(?) shapes plot instead of having a line graph.

middleware/middleware/api/project_api.py
SnowNooDLe: Good point, I could actually just overwrite instead of creating a new dict and copy.

claudio525 (approved):

jasonmotha (approved):

claudio525 (approved):

PR #71, opened 2021-04-22
Hazard curve percentiles

joelridden:
### Adding the 16th and 84th percentiles to the Hazard Curves
Sending an additional parameter to the core API calc_percentiles which is always set to true for the front end.
Currently only for non project Hazard Curves.
image
image


frontend/src/components/common/HazardCurve/HazardEnsemblePlot.js
```diff
@@ -71,6 +74,24 @@ const HazardEnsemblePlot = ({
...
+          },
+          // Percentile 84
+          {
+            x: percentile84.index,
+            y: percentile84.values,
+            type: "scatter",
+            mode: "lines",
+            name: "84th Percentile",
+            line: { color: "black", dash: "dot" },
+          },
```
SnowNooDLe, line 94: I am happy to go through but just realized that NZ Code is also using black dotted plots (Both NZTA and NZS1170p5 which isn't available yet until the PR goes through)

Thoughts? @claudio525 @jasonmotha

middleware/middleware/utils.py
```diff
@@ -65,7 +65,7 @@ def proxy_to_api(
         if "Download" in action:
             secret_key = (
                 DOWNLOAD_URL_SECRET_KEY_CORE
-                if "psha-core" in api_destination
+                if "psha-core" in api_destination or "10022" in api_destination
```
SnowNooDLe, line 68: @jasonmotha @claudio525 In case you wonder why I added this.

Depends on whether its a download action via CoreAPI/ProjectAPI, need to use the different secret_key 
However, if we need to test the Core/ProjectAPI that are running locally, `psha-core` will not be there, hence, with our conventions, checking with port number.

SnowNooDLe (dismissed):
Well done!

As I left a comment for Jason & Claudio, let's see what they think about having another dotted line along with NZ code plots(NZTA & NZS1170.5).

jasonmotha (approved):

SnowNooDLe (approved):

claudio525 (approved):

PR #73, opened 2021-04-28
Feedback update for students (EA mainly)

SnowNooDLe:
# Feedback update for students (EA mainly)

The following changes were made:
1. Vs30 dropdowns with the unit
2. Using subscript text for Vs30 in Regional/Vs30 tabs && headers to support images
3. Students will not have an access to GMS for now
    - Permission with a feature, not a specific role
4. Projects do have NZTA so, in theory, there will be another plot for NZTA, we just need to regenerate data
5. The way plots display the coordinates

## Permission updates - 30/04/2021

Just like we have for `hazard`, `project` now also has something like, `project:hazard`, `project:disagg`, `project:uhs` and `project:gms`. And by having feature permission, we can create many different roles, instead of specifying new permission for a certain role. (Thanks for clarifying/advice @claudio525)

### Example with my dummy account.
image

#### Non-projects, only `Hazard:hazard` and `Hazard:uhs` permission

image

#### Projects, only `Hazard:hazard` and `Hazard:uhs` permission

image

## HOTFIX - 30/04/2021

The following code is added to set the legends, the RP with its values(either site-specific or NZS1170.5).

```javascript
/*
        Only display to legend the RP that has values
      */
      const dataToCheck = isNZCode === true ? nzs1170p5Data : uhsData;
      for (let i = 0; i < selectedRPs.length; i++) {
        if (
          !Object.values(
            dataToCheck[`${1 / Number(selectedRPs[i]).toString()}`]
          ).includes("nan")
        ) {
          newLabel += `${selectedRPs[i].toString()}, `;
        }
      }
```
For instance, when RP is 4975, NZS1170.5 does not exist, hence not displaying RP 4975 in the legends.

image

image

## Proper Legendsgroup

Site-specific legends did not have a group setting properly, so when a user clicked site-specific legend, would only disable one of the plots. It's now fixed and legends will control every plot that is belonging to a certain legend group. (For this case, either nzs1170.5 or site-specific)

image

image

## Vs30 dropdown with the unit

image

## Using subscript text for Vs30 in Regional/Vs30 tabs && headers to support images

image

## Students will not have an access to GMS for now
With a new role called `ea-student`(Auth0) that contain `is-student` to check whether the user is a student, if so, the GMS tab is disabled.

image

## Projects do have NZTA so, in theory, there will be another plot for NZTA, we just need to regenerate data

image

## The way plots display the coordinates

Instead of displaying every y coordinates for a certain x coordinate, it now only displays the x&y coordinates that are close to the mouse cursor as plots are more like to be there to show the shapes instead of trying to find a certain value at a certain x or y coordinate.

### Hazard Curve

Screenshot 2021-04-28 15:37:31

### UHS

Screenshot 2021-04-28 15:36:32


jasonmotha (dismissed):

joelridden (dismissed):

CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
+## April 28, 2021
+
+### Feedback update for students (EA mainly) - ([PR #73](https://github.com/ucgmsim/seistech_psha_frontend/pull/73))
+
+1. Vs30 dropdowns with the unit.
+2. Using subscript text for Vs30 in Regional/Vs30 tabs && headers to support images.
+3. Students will not have an access to GMS for now.
```
claudio525, line 7: Was there any actual code change for this? Isn't that handled via Auth0, in which case I don't think it should be in the Changelog?

claudio525 (approved):

CHANGELOG.md
SnowNooDLe: @claudio525 Yes and No, Auth0 deals with permission name and with that permission, the frontend disables the tab
Inside `/views/Projects.js`
```javascript
disabled={validateTab() || hasPermission("is-student")}
```

SnowNooDLe: So that way, we can still access to GMS but students

jasonmotha (approved):

joelridden (approved):

joelridden (approved):

CHANGELOG.md
claudio525: @SnowNooDLe Lets have a chat regarding this when I'm in the office this arvo

jasonmotha (approved):

jasonmotha (approved):

jasonmotha (approved):

jasonmotha (approved):

frontend/src/components/Hazard/SiteSelection/SiteSelectionVS30.js
```diff
@@ -99,10 +104,12 @@ const SiteSelectionVS30 = () => {
       {vs30Map !== null &&
         showSpinner === false &&
         showErrorMessage.isError === false && (
-          <img
-            className="rounded mx-auto d-block img-fluid"
-            src={`data:image/png;base64,${vs30Map}`}
-            alt="Vs30 Map"
+          <ImageMap
+            header={
+              "Looking at the Foster et al. model predictions in the vicinity of the site."
```
claudio525, line 109: Maybe something like `The map below shows the Vs30 predictions in the vicinity of the site of interest, as given by the Foster et al. (2019) model`?

claudio525: Or shorter version: `Vs30 predictions in the vicinity of the site, as given by the Foster et al. (2019) model`

claudio525 (approved):

joelridden (approved):

PR #78, opened 2021-05-06
Metadata project id label

SnowNooDLe:
# Metadata project id label

### Before

image

### After

image


jasonmotha (approved):

frontend/src/components/Project/SeismicHazard/HazardCurveMetadata.js
```diff
@@ -29,7 +29,7 @@ const HazardCurveMetadata = ({
         */
         rows={nzCode.nztaSoil === null ? "8" : "9"}
         value={
-          `Project ID: ${projectId}\n` +
+          `Project ID: ${projectId["label"]}\n` +
```
claudio525, line 32: Nitpicking, but that should be called `Project Name` since its no longer the actual ID?
Also I think you should just have both, `Project Name` and `Project ID`. 
Metadata does not have to be pretty imo.

claudio525 (approved):

frontend/src/components/Project/SeismicHazard/HazardCurveMetadata.js
SnowNooDLe: Fair call haha
However, that `projectId` is something that looks like
```javascript
{
   value: "soffitel_queenstown",
   label: "Soffitel, Queenstown"
}
```

It's an object that has both `Project Name` and `Project ID`
I could add both `Project Name` and `Project ID`, one of the reasons I made the change was,

I chose `Soffitel, Queenstown` not `soffitel_queenstown` but suddenly the metadata is telling me something with `soffitel_queenstown` not the one I chose, `Soffitel, Queenstown` it that makes sense.

frontend/src/components/common/ProjectSelect.js
```diff
@@ -40,7 +40,7 @@ const ProjectSelect = ({
         id={id}
         value={value === undefined || value === null ? [] : value.value}
         placeholder={localOptions.length === 0 ? placeholder : "Select..."}
-        onChange={(e) => setSelect(e.value)}
+        onChange={(e) => isProjectID === true ? setSelect(e) : setSelect(e.value)}
```
joelridden, line 43: Should this be changed to something similar to what we changed for the UHS branches and percentiles?
Changing the value to be a local variable and then using the (projectIDOption) => setLocalSelectedProjectID(projectIDOption)

SnowNooDLe: Good catch.

As discussed after the morning stand up (11/05/2021), we are good to go :+1:

joelridden (approved):

PR #87, opened 2021-05-19
Split IM component

SnowNooDLe:
# Split IM component

Instead of having two different components for IM selector:
1. Projects
2. Non-projects

Now there is only one component that can be shared between two different tabs.

## Minor change in workflow

When a user chooses an option from the dropdown, the app used to store its value.
For instance, the options look like
```javascript
/* 
   Used to store `optionA` when a user clicks `Option A` label in the dropdown
   The app now stores an object, `{label: "Option A", value: "optionA"}`
*/
const options = [
       {label: "Option A", value: "optionA"},
       {label: "Option B", value: "optionB"}
]
```
However, to achieve resetting `react-select` component's value to `N/A` when the chosen IM is not `pSA`, I had to set an actual object by itself instead of using its value only.

Fortunately, not many changes were made as I could set the global variable with a `value` key instead of storing an actual object.

### Old

image

### New

image

image

image


frontend/src/components/Hazard/GMS/GMSForm.js
```diff
@@ -51,58 +54,61 @@ const GMSForm = () => {
...
-  const [localSFMin, setLocalSFMin] = useState(CONSTANTS.DEFAULT_SF_MIN);
-  const [localSFMax, setLocalSFMax] = useState(CONSTANTS.DEFAULT_SF_MAX);
-
-  const [localIMExdRateRadio, setLocalImExdRateRadio] = useState("im-level");
-
-  const [localIMLevel, setLocalIMLevel] = useState("");
-
-  const [localExcdRate, setLocalExcdRate] = useState("");
-
+  // For Caul params bounds data fetcher
```
jasonmotha, line 99: causal?

frontend/src/components/Hazard/GMS/GMSForm.js
```diff
@@ -51,58 +54,61 @@ const GMSForm = () => {
...
+  const [localIMVector, setLocalIMVector] = useState([]);
   const [localNumGMS, setLocalNumGMS] = useState("");
+  // Inputs under the Advanced tab
   const [localWeights, setLocalWeights] = useState({});
   const [localDatabase, setLocalDatabase] = useState({
     value: "nga_west_2",
     label: "nga_west_2",
   });
   const [localReplicates, setLocalReplicates] = useState(1);
+  // Causaul params bounds under the advanced
```
jasonmotha, line 87: causal*

jasonmotha (approved):

joelridden (dismissed):

joelridden (approved):

jasonmotha (approved):

claudio525 (approved):

PR #88, opened 2021-05-19
Tidy up Docker 

SnowNooDLe:
# Tidy up Docker

Tidy up the Docker not to install unnecessary packages and adapt it to install the Intermediate API from the target branch for testing.


jasonmotha (approved):

Jenkinsfile
```diff
@@ -1,63 +1,61 @@
-pipeline {
-    agent any
-    stages {
-        stage('Install dependencies') {
-            steps {
-                echo 'Install dependencies on Jenkins server (maybe unnecessary if test runs inside Docker)'
+// pipeline {
```
jasonmotha, line 1: Are we keeping this or deleting it?

SnowNooDLe: @jasonmotha I think ideally good to keep it as these will test the frontend but the current AWS instance cannot handle 9 containers simultaneously.

SnowNooDLe: Just disabled not to trigger to create 3 extra containers(Test purpose)

joelridden (dismissed):

Jenkinsfile
claudio525: I thought you wanted to re-do the frontend tests at some stage anyways? In which case we might as well use mantle at that point? So this can be removed?
Not really a fan of keeping code lying around that we might use one day (maybe if it's very complicated logic, but not sure if thats the case here)

claudio525: @SnowNooDLe

claudio525 (dismissed):

Jenkinsfile
SnowNooDLe: @claudio525 So this Jenkins is there to fire the Selenium(For UI testing), just like I do it locally while developing and at the same time testing was not our priority yet haven't really looked into the available tools to test. However, I am more than happy to remove test related files.

jasonmotha (approved):

joelridden (approved):

PR #93, opened 2021-06-21
Z1p0 z2p5 inputs

SnowNooDLe:
# Z<sub>1.0</sub> and Z<sub>2.5</sub> placeholders

image

The current site selection API call does not return any default values for Z values, unlike Vs30.

Hence, these inputs are there as placeholders and will be updated once we get
1. A default Z<sub>1.0</sub> and Z<sub>2.5</sub> values
2. Actually using Z<sub>1.0</sub> and Z<sub>2.5</sub> values

## Some unused variables
It's because these Z<sub>1.0</sub> and Z<sub>2.5</sub> are placeholders. We don't have any API calls for Z values yet.
Once we have implemented them, these variables will be used.
(E.g., APIQueryBuilder, handleErrors function, localSetZ1p0Click.... etc)

## TODO: Updae GMS plots to adopt the new JSON format
`metadata["rrup_error_bounds"]` is not longer a thing.

image


jasonmotha (approved):
All the python code is perfect. :smile:

joelridden (approved):

claudio525 (approved):

PR #96, opened 2021-07-13
HOTFIX - Permission Dashboards

SnowNooDLe:
# HOTFIX - Permission Dashboards

This issue happened while I was playing with my dummy account which gets created and removed frequently.
Solved by removing rows from Users_Permissions and Users_Projects table with a certain user_id that does not exist in Auth0 DB anymore.

#### More would come by working through the files but just added to db.py file for now as it's the only file that contains functions with some parameters.


middleware/middleware/api/intermediate_api.py
```diff
@@ -139,7 +139,8 @@ def get_all_users_projects():
     app.logger.info(
         f"Received request at {const.INTERMEDIATE_API_ALL_USERS_PROJECTS_ENDPOINT}"
     )
-    return db.get_all_users_project_permissions()
+    auth0_users, response_code = auth0.get_users()
```
joelridden, line 142: Do we use response_code? Could just make it a _

SnowNooDLe: It's not being used, is that `_` for an unused variable?

middleware/middleware/db.py
```diff
@@ -91,16 +96,25 @@ def _insert_permission(permission_name):
...
-def _is_user_access_permission_in_db(user_id, permission):
-    """Check whether there is a row with given user_id & permission"""
+def _is_user_access_permission_in_db(user_id: str, permission: str):
+    """Check whether there is a row with given user_id & permission
+    
+    Parameters
+    ----------
+    user_id: string
+        Auth0's unique ID
+    permission: str
```
joelridden, line 106: string instead of str? Just so its consistent

middleware/middleware/api/intermediate_api.py
joelridden: Yea we have done that a few times in Seistech

middleware/middleware/db.py
SnowNooDLe: used `str` instead of a `string` as If I were trying to replace `str` to `string`, had to do manually as parameter's data type is also str :joy:
Also, noticed that we use `str` in the seistech repo's docstring.

middleware/middleware/db.py
```diff
@@ -281,10 +315,36 @@ def get_all_users_project_permissions():
...
+    with Users who are no longer on board
+
+    Parameters
+    ----------
+    auth0_users: Dict
+        All the users who are signed up for PSHA app
+    """
+    all_users_in_users_projects = list(
+        set([user.user_id for user in models.UserProject.query.all()])
+    )
```
joelridden, line 330: I think you can do set comprehension here to reduce the amount of conversions
`all_users_in_users_projects = list({user.user_id for user in models.UserProject.query.all()})`

SnowNooDLe: Sweet, `{}` keeps me only thinking of a dictionary, not a set :joy:

middleware/middleware/db.py
```diff
@@ -305,7 +367,28 @@ def get_all_users_permissions():
...
+    with Users who are no longer on board
+
+    Parameters
+    ----------
+    auth0_users: Dict
+        All the users who are signed up for PSHA app
+    """
+    all_users_in_users_permissions = list(
+        set([user.user_id for user in models.UserPermission.query.all()])
+    )
```
joelridden, line 382: Same thing here.
These functions are also rather similar, not sure if it'd be worth though making one function where you pass in the model to query and modify as they are rather short functions anyway.

SnowNooDLe: Yeah, two places but create a function for this simple, wasn't sure whether it would be worth doing haha.
```python
return list({user.user_id for user in models.UserPermission.query.all()})
```

jasonmotha (dismissed):

joelridden (dismissed):

middleware/middleware/db.py
```diff
@@ -281,10 +315,28 @@ def get_all_users_project_permissions():
     return users_projects_dict
 
 
-def get_all_users_permissions():
+def _sync_users_projects_db(auth0_users: Dict):
```
claudio525, line 318: Why not use `_sync_users_table_with_auth0` directly instead of creating a new function?

middleware/middleware/db.py
```diff
@@ -305,12 +359,25 @@ def get_all_users_permissions():
     return users_permissions_dict
 
 
-def allocate_projects_to_user(user_id, project_list):
+def _sync_users_permissions_db(auth0_users: Dict):
```
claudio525, line 362: Same with this one?

middleware/middleware/db.py
SnowNooDLe: Great question and I did it because of the naming thing.
E.g., you look at the function name with `get_all_users_project_permissions()` and the first thing it does is sync user_id between Users_projects table and Auth0's user id which is that `_sync_users_projects_db()` for whereas `_sync_users_table_with_auth0()` could end up a minor confusion with Users table we have instead of Users_Projects which is the table we need to sync.

claudio525: What about just having one function that does all the syncing? I.e. it syncs user project and user permissions?

jasonmotha (approved):

joelridden (approved):

claudio525 (approved):

PR #101, opened 2021-07-21
Minor bug updates

SnowNooDLe:
# Minor bug updates

- Rename the custom IM Select component name to `IMCustomSelect.js`.
- Fix to set the selected value when there is only one option for the Component dropdown.
  - Both non-projects and projects
  - May extend this feature to any other dropdowns with only one possible option but keep it with Component for now.
- Hazard Curve **compute** button stays disabled if a component is not selected.
- Disaggregation **compute** button stays disabled if a component is not selected.
- Update the **Compute** button logic on the NZ Code section
  - If only Hazard Curve got computed, depends on the selected IM, the button gets disabled or enabled(Only with PSA and pSA)
  - If only UHS got computed, the button gets enabled as UHS doesn't need any IM value.
  - If both got computed, the button gets enabled regardless of the selected IM. However, based on the selected IM, the app may update NZ Codes for both Hazard Curve and UHS or just UHS.
- Project's Hazard Curve **compute** button stays disabled if a component is not selected.
- Project's Disaggregation **get** button stays disabled if a component is not selected.

## Compute button's logic - Disabled if a component is not selected
image
image


frontend/src/components/Hazard/SeismicHazard/HazardCurveSection.js
```diff
@@ -106,11 +112,7 @@ const HazardCurveSection = () => {
           id="im-select"
           type="button"
           className="btn btn-primary"
-          disabled={
-            localSelectedIM === null ||
-            (localSelectedIM["value"] === "pSA" &&
-              localSelectedIMPeriod === null)
-          }
+          disabled={validateInputs()}
```
joelridden, line 115: Maybe we should have this as !validateInputs() like we have done for the Disagg Section and just flip the logic for valid inputs

frontend/src/components/Project/SeismicHazard/DisaggregationSection.js
```diff
@@ -53,11 +66,7 @@ const DisaggregationSection = () => {
           id="project-hazard-curve-get-btn"
           type="button"
           className="btn btn-primary"
-          disabled={
-            localSelectedRP === null ||
-            projectSelectedIM === null ||
-            (projectSelectedIM === "pSA" && projectSelectedIMPeriod === null)
-          }
+          disabled={!validInputsNonpSA() || validInputWithpSA()}
```
joelridden, line 69: Maybe a rename to invalidInputWithpSA?

frontend/src/components/Hazard/SeismicHazard/HazardCurveSection.js
SnowNooDLe: @joelridden I will just go over all the `validators` we have for buttons and rename them a little bit.
For instance, `invalidInputs` like you suggested in another comment.

Cuz looking at this line seems odd as if it's valid, we disable the button :thinking: 
```javascript
disabled={validateInputs()} 
```

joelridden (approved):

claudio525 (dismissed):

jasonmotha (dismissed):

SnowNooDLe:
The comment was about a typo and everyone approved the code, merged the branch.

PR #104, opened 2021-08-04
Scenarios

joelridden:
### **Adding Scenarios to Projects and non-projects**

- New Tab addition with new access permissions (hazard:scenarios, project:scenarios)

image
User selects an IM component then can hit Compute or Get depending on if the user is in Non-projects or Projects.
Then the user can select different scenarios/ruptures from the dropdown to show on the plot, currently 16,50 and 84th percentiles are shown on the plot with the same colour as the scenario and different per scenario.


frontend/src/components/Hazard/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,128 @@
...
+import Select from "react-select";
+import makeAnimated from "react-select/animated";
+
+import * as CONSTANTS from "constants/Constants";
+import { GlobalContext } from "context";
+
+import { IMCustomSelect, GuideTooltip } from "components/common";
+import { createSelectArray } from "utils/Utils";
+
+import "assets/style/GMSForm.css";
```
SnowNooDLe, line 13: Do we need this css for ScenarioForm.js?

frontend/src/components/Hazard/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,128 @@
...
+    setScenarioData(null);
+  }, [station]);
+
+  return (
+    <Fragment>
+      <div className="form-group form-section-title">
+        Scenarios
+        <GuideTooltip
+          explanation={CONSTANTS.TOOLTIP_MESSAGES["HAZARD_HAZARD"]}
+        />
```
SnowNooDLe, line 76: Might be better to leave TODO comment for the future so we can ask for the right guide message for Scenarios.

frontend/src/components/Hazard/Scenarios/ScenarioViewer.js
```diff
@@ -0,0 +1,165 @@
...
+        scenarioData !== null &&
+        showErrorMessage.isError === false && (
+          <Fragment>
+            <ScenarioPlot
+              scenarioData={scenarioData}
+              scenarioSelectedRuptures={scenarioSelectedRuptures}
+              extra={extraInfo}
+            />
+            <DownloadButton
+              disabled={downloadToken === ""}
```
SnowNooDLe, line 152: For this state, do we want to reset to an empty string while fetching?
LIke state updaters between line number 62~64.

I don't think there will be any case that token is different to the current plots' data as unlike any other viewers, seems this `DownloadButton` only gets visible when plot is showing?

frontend/src/components/Project/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,132 @@
...
+import Select from "react-select";
+import makeAnimated from "react-select/animated";
+
+import * as CONSTANTS from "constants/Constants";
+import { GlobalContext } from "context";
+
+import { IMCustomSelect, GuideTooltip } from "components/common";
+import { createSelectArray } from "utils/Utils";
+
+import "assets/style/GMSForm.css";
```
SnowNooDLe, line 13: Same to this one

frontend/src/components/Project/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,132 @@
...
+    setProjectScenarioData(null);
+  }, [projectId, projectVS30, projectLocation, projectZ1p0, projectZ2p5]);
+
+  return (
+    <Fragment>
+      <div className="form-group form-section-title">
+        Scenarios
+        <GuideTooltip
+          explanation={CONSTANTS.TOOLTIP_MESSAGES["HAZARD_HAZARD"]}
+        />
```
SnowNooDLe, line 80: Possible TODO comment I reckon.

frontend/src/components/Project/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,132 @@
...
+          setSelect={setLocalSelectedIMComponent}
+          options={projectScenarioIMComponentOptions}
+          selectedIM={"pSA"}
+          placeholder={"Loading..."}
+        />
+      </div>
+
+      <div className="form-group">
+        <button
+          id="im-select"
```
SnowNooDLe, line 94: It's not an end of the world but probably better to remind both of us that `id` should/must be unique else just use `class`.
I know that I also made the same mistakes but we probably should follow what these attributes ask us for.  :joy:

frontend/src/components/Project/Scenarios/ScenarioForm.js
```diff
@@ -0,0 +1,132 @@
...
+          disabled={localSelectedIMComponent === null}
+          onClick={() => setProjectScenarioGetClick(uuidv4())}
+        >
+          Compute
+        </button>
+      </div>
+
+      <div className="form-group">
+        <label
+          id="label-ruptures"
```
SnowNooDLe, line 106: Same comment to the one above.

middleware/middleware/api/core_api.py
```diff
@@ -348,6 +348,14 @@ def get_gms_available_ims():
     )
 
 
+# Scenarios
+@app.route(const.CORE_API_SCENARIOS_ENDPOINT, methods=["GET"])
+def get_scenario():
+    return utils.proxy_to_api(
+        request, const.ENSEMBLE_SCENARIO_ENDPOINT, "GET", CORE_API_BASE, CORE_API_TOKEN
```
SnowNooDLe, line 355: Shall we provide the last two params? `user_id` and `action` to store their activity like we did in Project's scenarios get

frontend/src/components/common/Scenarios/ScenarioPlot.js
```diff
@@ -0,0 +1,146 @@
...
+    const x_values = [];
+    for (const key in ims) {
+      if (ims[key] === "PGA") {
+        x_values.push("0");
+      } else if (ims[key].includes("pSA")) {
+        x_values.push(ims[key].split("_")[1]);
+      } else {
+        x_values.push(ims[key]);
+      }
+    }
```
SnowNooDLe, line 40: We can also do something like this

```javascript
const test = ["PGA", "pSA", "PAA"];

    let temp = []

    test.forEach((IM) => {
        if (IM === "PGA") temp.push(IM)
        else if (IM === "pSA") temp.push("I WAS PSA")
        else temp.push("oW!")
    })
```

SnowNooDLe (commented):
Some minor comments from my end, otherwise good job!

frontend/src/components/Hazard/Scenarios/ScenarioViewer.js
joelridden: Fair point the downloadToken shoudl be set when the plot is showing so this can be removed yea

frontend/src/components/common/Scenarios/ScenarioPlot.js
joelridden: I like this

claudio525:
@joelridden (@joelridden ) This PR should be into `master` not `early_access` (can discuss tomorrow)

jasonmotha (approved):
Happy in general.

claudio525 (approved):

SnowNooDLe (approved):

jasonmotha (approved):

